PR #1, opened 2016-04-08
[JENKINS-26107] Labeled block

jglick:
Moved from https://github.com/jenkinsci/pipeline-plugin/pull/140.

[JENKINS-26107](https://issues.jenkins-ci.org/browse/JENKINS-26107)

- [X] basic impl
- [X] test
- [X] fix `BodyInvoker.displayName`: https://github.com/jenkinsci/workflow-cps-plugin/pull/3
- [x] review appearance in _Pipeline Steps_

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/LabelStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/LabelStep.java
@@ -1,0 +1,80 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 Jesse Glick.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.steps;
+
+import hudson.Extension;
+import javax.inject.Inject;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class LabelStep extends AbstractStepImpl {
+
+    private final String name;
+
+    @DataBoundConstructor public LabelStep(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public static class Execution extends AbstractStepExecutionImpl {
+
+        private static final long serialVersionUID = 1L;
+        
+        @Inject private transient LabelStep step;
+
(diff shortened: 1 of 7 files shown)


jglick:
@HRMPW notes that the term _label_ is also used to classify and select agents, for example as a parameter to the `node` step, so its use here might be confusing. Is there an alternate term which would be intuitive as a step name, and not overloaded in a Jenkins context? _segment_?

src/main/java/org/jenkinsci/plugins/workflow/steps/LabelStep.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 Jesse Glick.
```
amuniz: 2016?

amuniz:
:bee: - regardless the final name for this step.

emanuelez:
FWIW I think the name `label` fits fine and I don't see a lot of confusion happening with node labels.
Looking forward to this! :)

abayer:
I'd vote for segment or section or something like that over an already meaningful term like label.

svanoort:
@jglick Is there a reason we don't attach a LabelAction here?  

Personally I like calling this a 'Tag' or 'NamedBlock.'   

Point of note: I am realizing that there are some serious complexities in this and in mixing and matching stages and $WhateverWeCallThis.   Stages are ordered by either iota or startTime and unscoped, but $ThisThing is scoped.   I think we should also define how we expect those 'what contains what' issues to be handled.  

A couple things to consider:

* Due to the fact that this is attached to blocks (which can nest), we may nest labelled blocks, and indeed may mix named sub-blocks with atomic steps, i.e.:

```java
label('top') {
  sh 'cp myFile newFile'
  label('more') {
    sh 'bash doMore.sh'  // Nested
      label('or_less') {  // Nested under or_less, more, then top
        sh 'less newFile'
      }
  }
  sh 'echo "we are done"'  // Nested under top
}
echo 'But wait there is more!'  //I have no label?
```

* Evaluating whether a step executed as part of a stage or a labelled block is rather complex: 
``java
stage 'mine'
echo 'I am part of the stage.'
label('no, mine!') {
  echo 'Clearly, I belong to the label.  Maybe also the stage?'
}
echo 'but who owns me? stage, presumably.'
```

Let's not even consider the case where we have a stage preceding a parallel block, with labels mixed in.

We will obviously have to store the scopes of blocks in a stage as we walk the Flow graphs to evaluate where steps ran (not a big deal), and also the list of stage steps (to see if something happened before or during one). 

Consider also: do we want to describe labels/tags/names as a *list* with all applicable values, from the outside in?  No change needed in the step itself, I think, but only in how we evaluate it within UIs. This may actually be more useful and logical for users than seeing (for example) the nearest descriptive value.  In addition it isn't the end of the world if the ordering is off sometimes.   It also avoids the puzzling behavior (for users) of seeing that last echo step as part of a stage that doesn't appear to contain the preceding blocks (since they have separate labels applied). 

Maybe I'm missing something in the intended use here?

**Comment by jglick** [comment]
> Is there a reason we don't attach a LabelAction here?

Huh? We do.

> Let's not even consider the case where we have a stage preceding a parallel block, with labels mixed in.

The intent in that case is clear to my mind. The code between one `stage` and the next serves as an implicit labeled block, enclosing the explicit ones. `FlowNodeSerialWalker` already implemented this.

The tricky case is `stage` inside some (non-`parallel`) labeled block; for that case I think we can simply say the behavior is undefined and leave it at that, since `stage` was only ever intended for top-level use.

Anyway with the advent of `label` you should just stop using `stage` anyway. There is no need to support scripts using both.

```groovy
node {
  stage 'Check out'
  checkout scm
  stage 'Build'
  sh 'make'
  stash name: 'app', includes: 'app.bin'
}
node {
  stage 'Test'
  unstash 'so'
  sh './app.bin'
}
```

should be translated to something like

```groovy
label('Dev') {
  node {
    label('Check out') {
      checkout scm
    }
    label('Build') {
      sh 'make'
      stash name: 'app', includes: 'app.bin'
    }
  }
}
label('Test') {
  node {
    unstash 'so'
    sh './app.bin'
  }
}
```

You could choose to leave the `node` blocks outside of `label`, in which case they would simply have no associated label, so any messages or pauses or metadata (for example, waiting for a free executor) would just be show at top level in some UI-specific manner, as if the entire build were wrapped in an implicit

```groovy
label('<top level>') {
  // …
}
```

> This may actually be more useful and logical for users than seeing (for example) the nearest descriptive value.

No particular opinion. Different UIs may have different needs, and different amount of space available.

jglick:
@kohsuke wants this as a mode of `stage`.

abayer:
+1 to that. =)

svanoort:
It certainly reduces user confusion, but I'm wincing a bit thinking of how we'll handle pipelines with two modes of operation for stage (snapshot-in-time vs. block-scoped).

PR #2, opened 2016-04-13
[JENKINS-28119] ErrorInfoStep

jglick:
[JENKINS-28119](https://issues.jenkins-ci.org/browse/JENKINS-28119)

- [X] prototype
- [ ] tail of log
- [ ] label
- [ ] fuller test case

@reviewbybees

PR #3, opened 2016-04-27
Flowscanner reimplement jenkins28119 error info

svanoort:
Modifies https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/2 to use the new flow scanners, supplying predicates. 

@jglick This is the first of the implementations with the new FlowScanners.   Would be happy to resubmit as a PR against your fork if you prefer.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,12 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>1.15</version>
+            <version>2.1-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>script-security</artifactId>
+            <version>1.18</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -90,7 +90,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>1.15</version>
+            <version>2.2-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/ErrorInfoStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/ErrorInfoStep.java
@@ -60,8 +60,9 @@ public class ErrorInfoStep extends AbstractStepImpl {
 
     public static class Execution extends AbstractSynchronousStepExecution<ErrorInfo> {
 
-        @Inject private ErrorInfoStep step;
-        @StepContextParameter private FlowExecution execution;
+        private static final long serialVersionUID = 1;
+        @Inject private transient ErrorInfoStep step;
+        @StepContextParameter private transient FlowExecution execution;
 
         @Override protected ErrorInfo run() throws Exception {
             return new ErrorInfo(step.error, execution);
@@ -174,7 +175,7 @@ public class ErrorInfoStep extends AbstractStepImpl {
                 for (FlowNode n2 : new FlowNodeSerialWalker(n)) {
                     LogAction a = n2.getAction(LogAction.class);
                     if (a != null) {
-                        String u = Jenkins.getInstance().getRootUrl();
+                        String u = Jenkins.getActiveInstance().getRootUrl();
                         if (u == null) {
                             u = "http://jenkins/"; // placeholder
                         }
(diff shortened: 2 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/ErrorInfoStep.java
```diff
@@ -0,0 +1,193 @@
...
+                @Override
+                public boolean apply(FlowNode input) {
+                    if (input instanceof BlockEndNode) {
+                        return false;
+                    }
+                    ErrorAction a = input.getAction(ErrorAction.class);
+                    return (a != null && stackTraces.contains(Functions.printThrowable(a.getError())));
+                }
+            };
+            return new FlowScanner.DepthFirstScanner().findFirstMatch(getExecution(), threwException);
```
svanoort: Once the ForkScanner is complete, it should be far more efficient for this case, because:

1. It doesn't have to keep a HashSet of all visited FlowNodes (just a stack with depth related to the number of levels of parallel branches). 
2. DepthFirstScanner will walk to the start of the flow, if the failure occurs in any branch branch of a parallel block beyond the first.  ForkScanner will only walk the up to the start of the parallel block containing the failure (and then go through the branches there). 

Technically the completely optimal case here is probably to skip blocks without an ErrorAction (if we can guarantee they're attached to all the BlockEndNodes where failures happen) AND to do a proper breadth-first search.

src/main/java/org/jenkinsci/plugins/workflow/steps/ErrorInfoStep.java
```diff
@@ -0,0 +1,193 @@
...
+         * Note that {@link Throwable#equals} is just pointer equality,
+         * which we cannot use since we may be loading deserialized exceptions,
+         * so we compare by stack trace instead.
+         */
+        private @CheckForNull FlowNode getNode() throws IOException {
+            final Set<String> stackTraces = new HashSet<>();
+            for (Throwable t = error; t != null; t = t.getCause()) {
+                stackTraces.add(Functions.printThrowable(t));
+            }
+            Predicate<FlowNode> threwException = new Predicate<FlowNode>() {
```
jglick: I would inline this variable. Just make sure it is safe for lambda conversion when we switch to `-source 8`. Most likely this is satisfied by just verifying that there are no method overloads which vary only in the declared type of the SAM they accept. (This is where we screwed up with `ACL.impersonate`.)

src/main/java/org/jenkinsci/plugins/workflow/steps/ErrorInfoStep.java
jglick: `DepthFirstScanner` is a top-level type in the current upstream PR.

For a PR downstream of an evolving API PR, it is best to use timestamped snapshots in the Maven dependency, rather than `2.1-SNAPSHOT` which is vague.

pom.xml
```diff
@@ -85,12 +85,17 @@
...
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>script-security</artifactId>
+            <version>1.18</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>1.15</version>
+            <version>2.2-SNAPSHOT</version>
```
jglick: BTW this can now be switched to simply `2.2`.

src/main/java/org/jenkinsci/plugins/workflow/steps/ErrorInfoStep.java
jglick: Performance is not a significant concern for this PR, because this is code run just once per step execution, typically once per build, and from the CPS VM thread. Obviously if you have some system for indexing nodes using specific action classes, that would be helpful as you could go straight to those with `ErrorAction`.

I did not follow the discussion about `ForkScanner` here. At any rate, for purposes of this PR we are looking for any `FlowNode` in the build with an `ErrorAction`, so long as it is not a `BlockEndNode`. Possibly we really want to check also `BlockEndNode`s, to handle errors thrown only during cleanup from build wrappers and the like, but give them a lower priority. The issue is that `ErrorAction` gets attached to end nodes of blocks which are exited by an uncaught exception, so we want to find the “first” node with the matching action: i.e., giving preference to the nested node. So it would also suffice to do a walk forward in time, rather than backward as I presume `DepthFirstScanner` does. Seems there is no such scanner yet?

src/main/java/org/jenkinsci/plugins/workflow/steps/ErrorInfoStep.java
```diff
@@ -0,0 +1,193 @@
...
+                        u = "http://jenkins/"; // placeholder
+                    }
+                    return u + logNode.getUrl() + logNode.getAction(LogAction.class).getUrlName();
+                }
+            }
+            return null;
+        }
+
+        // TODO tail of log
+        // TODO label
```
jglick: So what would this look like? (Searching for `LabelAction` and/or `StageAction` as per `FlowNodeSerialWalker`’s convenience API for labels.)

PR #5, opened 2016-05-12
Clarify behaviour of retry step in help and fix English on timeout step help

rgal:
This answers a question I had in trying to use retry, whether it would actually cause an exception if the block continued to fail.

I looked in the source code and tried this out to confirm that this is what retry does but it would be useful to say it in the help.


jglick:
> whether it would actually cause an exception if the block continued to fail

Yes, the exception would just be thrown up, as if there had been no `retry` block.

PR #7, opened 2016-05-20
[FIXED JENKINS-34554] WaitForConditionStep: introduce maximum recurrence period

sonneveld:
Implements [JENKINS-34554](https://issues.jenkins-ci.org/browse/JENKINS-34554).  Wasn't entirely sure how to implement a test (or if one was necessary).


jglick:
A configurable limit as in JENKINS-29037 would probably be nice, but this seems a reasonable first step.

SmallConcern:
Is there a way to modify this value in the waitUntil block or configure it some other way? A max recurrence time of 15s seems a little too fast for some polling needs.

dcworldwide:
Yes is this configurable? I couldn't find any params in the docs https://jenkins.io/doc/pipeline/steps/workflow-basic-steps/#code-waituntil-code-wait-for-condition

stefanteixeira:
15 seconds is too little time for some polling needs, why not letting users configure this according to their needs? 😕

dwnusbaum:
@stefanteixeira Feel free to file a PR that makes it configurable via a new optional parameter.

PR #8, opened 2016-06-08
[JENKINS-34281] Indicate if we are still sleeping after a resume

jglick:
Related to https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/4: I noticed that `sleep` steps were sometimes hanging. Turns out that was another bug, to be filed soon…

@reviewbybees


svanoort:
🐝

christ66:
🐝

PR #11, opened 2016-07-04
Workaround for BUG JENKINS-15146; support for empty environment variables

lindt:
```groovy
withEnv(['FOO=']) { }
```
should set the Environment Variable "FOO" to an empty String. But instead of this it completly removes the Environment Variable.


jglick:
The proposed fix is not acceptable I think; a space is not the same as the empty string.

JENKINS-15146 is not the right issue; that is about EnvInject. If there is a related bug reproducible in Pipeline, it should be filed separately.

And any patch like this requires a test demonstrating the bug and the effectiveness of the fix.

PR #12, opened 2016-07-08
JENKINS-29170 - `Util.fixEmpty` `charset` and `mimeType`

Evildethow:
[JENKINS-29170](https://issues.jenkins-ci.org/browse/JENKINS-29170)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
@@ -27,6 +27,7 @@ import com.google.inject.Inject;
 
 import hudson.AbortException;
 import hudson.Extension;
+import hudson.Util;
 import hudson.model.TaskListener;
 import jenkins.plugins.mailer.tasks.MimeMessageBuilder;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
```diff
@@ -145,10 +146,10 @@ private MimeMessage buildMimeMessage() throws UnsupportedEncodingException, Mess
             if (step.bcc != null) {
                 messageBuilder.addRecipients(step.bcc, Message.RecipientType.BCC);
             }
-            if (step.charset != null) {
+            if (Util.fixEmpty(step.charset) != null) {
```
jglick: :bug: do not do this here. Rather do it in the setters. Read the [guide](https://github.com/jenkinsci/pipeline-plugin/blob/master/DEVGUIDE.md#handling-default-values).

Evildethow:
@jglick Thanks for the pointer. Moved into setters to support the snippet generator

src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
```diff
@@ -50,7 +51,6 @@
  */
 public class MailStep extends AbstractStepImpl {
 
-    @DataBoundSetter
     public String charset;
```
jglick: May as well make them `private` while you are at it.

jglick:
:bee:

jglick:
re:bee:

PR #13, opened 2016-07-15
Update help.html

aurelije:
It is not clear from documentation that MINUTES are default unit


src/main/resources/org/jenkinsci/plugins/workflow/steps/TimeoutStep/help.html
```diff
@@ -2,4 +2,6 @@
     Executes the code inside the block with a determined time out limit.
     If the time limit is reached, an exception is thrown, which leads in aborting 
     the build (unless it is caught and processed somehow).
+    
+    If unit is omitted than the default time unit applied is MINUTES
```
jglick: typo

jglick:
Use `help-time.html` and `help-unit.html`.

dwnusbaum:
https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/59 updated the help text for the `timeout` step to indicate that the default unit is minutes, so I am closing this PR.

PR #14, opened 2016-07-15
Update help.html

aurelije:
Not directly clear what is the default time unit. One more problem is that sleep method exists in groovy also, but their time unit is MILLISECONDS and IDE will try to do code completion using that method and will complaint when different arguments are supplied to the pipeline method. A little bit confusing.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/steps/SleepStep/help.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/steps/SleepStep/help.html
@@ -2,4 +2,6 @@
     Simply pauses the Pipeline build until the given amount of time has expired.
     Equivalent to (on Unix) <code>sh 'sleep …'</code>.
     May be used to pause one branch of <code>parallel</code> while another proceeds.
+    
+    If unit is ommited then default time unit is SECONDS 
 </div>


src/main/resources/org/jenkinsci/plugins/workflow/steps/SleepStep/help.html
```diff
@@ -2,4 +2,6 @@
     Simply pauses the Pipeline build until the given amount of time has expired.
     Equivalent to (on Unix) <code>sh 'sleep …'</code>.
     May be used to pause one branch of <code>parallel</code> while another proceeds.
+    
+    If unit is ommited then default time unit is SECONDS
```
jglick: typo

jglick:
Would be better to have `help-fieldName.html` for each field.

jglick:
> sleep method exists in groovy also, but their time unit is MILLISECONDS

Yes; OTOH the POSIX `/usr/bin/sleep` (which I would hope more people are familiar with) uses seconds.

dwnusbaum:
See https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/114.

PR #16, opened 2016-08-14
[FIXED JENKINS-37397] Allow use of @Symbols from tools as tool types.

abayer:
[JENKINS-37397](https://issues.jenkins-ci.org/browse/JENKINS-37397)

Fall back to allowing descriptor IDs as well.

[JENKINS-37386](https://issues.jenkins-ci.org/browse/JENKINS-37386) is driving getting `@Symbol` on `ToolDescriptor`s, so this will not be hugely useful until that's everywhere, but still, better to do ahead of time than after the fact!

cc @reviewbybees esp @jglick and @oleg-nenashev

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/ToolStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/ToolStep.java
@@ -89,7 +89,7 @@ public final class ToolStep extends AbstractStepImpl {
             ListBoxModel r = new ListBoxModel();
             r.add("<any>", "");
             for (ToolDescriptor<?> desc : ToolInstallation.all()) {
-                String idOrSymbol = symbolForDescriptor(desc);
+                String idOrSymbol = SymbolLookup.getSymbolValue(desc);
                 if (idOrSymbol == null) {
                     idOrSymbol = desc.getId();
                 }
@@ -103,7 +103,7 @@ public final class ToolStep extends AbstractStepImpl {
             ListBoxModel r = new ListBoxModel();
 
             for (ToolDescriptor<?> desc : ToolInstallation.all()) {
-                if (type != null && !desc.getId().equals(type) && !type.equals(symbolForDescriptor(desc))) {
+                if (type != null && !desc.getId().equals(type) && !type.equals(SymbolLookup.getSymbolValue(desc))) {
                     continue;
                 }
                 for (ToolInstallation tool : desc.getInstallations()) {
@@ -126,7 +126,7 @@ public final class ToolStep extends AbstractStepImpl {
             String name = step.getName();
             String type = step.getType();
             for (ToolDescriptor<?> desc : ToolInstallation.all()) {
-                if (type != null && !desc.getId().equals(type) && !type.equals(symbolForDescriptor(desc))) {
+                if (type != null && !desc.getId().equals(type) && !type.equals(SymbolLookup.getSymbolValue(desc))) {
                     continue;
                 }
                 for (ToolInstallation tool : desc.getInstallations()) {
@@ -146,25 +146,6 @@ public final class ToolStep extends AbstractStepImpl {
 
         private static final long serialVersionUID = 1L;
 
-    }
-
-    /**
-     * Finds the {@code @Symbol} on the given {@link ToolDescriptor}, if it exists.
-     * TODO: This should probably go somewhere else - maybe {@link SymbolLookup}?
-     *
-     * @param desc A {@link ToolDescriptor}
-     * @return The {@code @Symbol} value for that descriptor, or null if not present.
-     */
-    public static String symbolForDescriptor(ToolDescriptor<?> desc) {
-        Symbol s = desc.getClass().getAnnotation(Symbol.class);
-        if (s != null) {
-            for (String t : s.value()) {
-                if (t != null) {
-                    return t;
-                }
-            }
-        }
-        return null;
     }
 
 }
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/ToolStep.java
```diff
@@ -119,7 +126,7 @@ public ListBoxModel doFillNameItems(@QueryParameter String type) {
             String name = step.getName();
             String type = step.getType();
             for (ToolDescriptor<?> desc : ToolInstallation.all()) {
-                if (type != null && !desc.getId().equals(type)) {
+                if (type != null && !desc.getId().equals(type) && !type.equals(SymbolLookup.getSymbolValue(desc))) {
```
jglick: Very hard to glance at this code and see that it is correct. Make sure you have test coverage for all applicable cases. Probably best to define some dummy `ToolInstallation` types with and without symbols and use `@TestExtension`.

jglick:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/steps/ToolStep.java
abayer: Good idea - I've verified it works both against 1.x and 2.x with Maven (i.e., without and with a symbol) but having an explicit test in there makes sense. I'll do that shortly.

src/test/java/org/jenkinsci/plugins/workflow/steps/ToolStepRunTest.java
```diff
@@ -60,4 +73,69 @@
...
+            super(name, home, properties);
+        }
+
+        @TestExtension
+        @Symbol("mockToolWithSymbol")
+        public static class MockToolWithSymbolDescriptor extends ToolDescriptor<MockToolWithSymbol> {
+            @Override
+            @Nonnull
+            public String getDisplayName() {
+                return "MockToolWithSymbol";
```
jglick: Can be deleted in this version of Jenkins—defaults to simple class name.

src/test/java/org/jenkinsci/plugins/workflow/steps/ToolStepRunTest.java
```diff
@@ -60,4 +73,69 @@
...
+        public MockToolWithoutSymbol(String name, String home, List<? extends ToolProperty<?>> properties) {
+            super(name, home, properties);
+        }
+
+        @TestExtension
+        public static class MockToolWithoutSymbolDescriptor extends ToolDescriptor<MockToolWithoutSymbol> {
+            @Override
+            @Nonnull
+            public String getDisplayName() {
+                return "MockToolWithoutSymbol";
```
jglick: ditto

src/test/java/org/jenkinsci/plugins/workflow/steps/ToolStepRunTest.java
```diff
@@ -60,4 +73,69 @@
         r.assertLogContains("Apache Maven 3", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
     }
 
+    @Test public void toolWithSymbol() throws Exception {
+        File buildDirectory = new File(System.getProperty("buildDirectory", "target")); // TODO relative path
```
jglick: :ant: either relativize to `r.jenkins.getRootDirectory()`, or use `@Rule public TemporaryFolder`.

src/test/java/org/jenkinsci/plugins/workflow/steps/ToolStepRunTest.java
```diff
@@ -60,4 +73,69 @@
...
+        File toolHome = new File(buildDirectory, "mockTools");
+        MockToolWithSymbol tool = new MockToolWithSymbol("mock-tool-with-symbol", toolHome.getAbsolutePath(), JenkinsRule.NO_PROPERTIES);
+        Jenkins.getInstance().getDescriptorByType(MockToolWithSymbol.MockToolWithSymbolDescriptor.class).setInstallations(tool);
+        String name = tool.getName();
+
+        String type = "mockToolWithSymbol";
```
jglick: Both of these are constants and so could be inlined into the script.

src/test/java/org/jenkinsci/plugins/workflow/steps/ToolStepRunTest.java
```diff
@@ -60,4 +73,69 @@
...
+
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node {def home = tool name: '" + name + "', type: '" + type + "'}",
```
jglick: A little stronger to `echo` that `home` to make sure it is `toolHome`.

src/test/java/org/jenkinsci/plugins/workflow/steps/ToolStepRunTest.java
```diff
@@ -60,4 +73,69 @@
...
+        Jenkins.getInstance().getDescriptorByType(MockToolWithoutSymbol.MockToolWithoutSymbolDescriptor.class).setInstallations(tool);
+        String name = tool.getName();
+
+        String type = "mockToolWithoutSymbol";
+
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node {def home = tool name: '" + name + "', type: '" + type + "'}",
+                true));
+
+        r.assertLogContains("No mockToolWithoutSymbol named mock-tool-without-symbol found",
```
jglick: Should probably test the _normal_ case that you can specify the descriptor ID and have it work!

jglick:
re:bee:

abayer:
@reviewbybees done

PR #17, opened 2016-08-29
[JENKINS-31931] Adding test case

jglick:
[JENKINS-31931](https://issues.jenkins-ci.org/browse/JENKINS-31931)

Also fixing a buglet that archiving from a nonexistent workspace should not automatically fail, except insofar as archiving no files would. Cf. [JENKINS-28121](https://issues.jenkins-ci.org/browse/JENKINS-28121) which also deals with nonexistent workspaces. Generally it is up to each step which expects a context directory to exist to actually make that directory; otherwise `dir('mayOrMayNotExist') {fileExists 'something'}` would gratuitously `mkdirs`, for example.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,8 +28,8 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.5</version>
-        <relativePath />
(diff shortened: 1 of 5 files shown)


svanoort:
🐝

PR #18, opened 2016-09-17
[JENKINS-26521] Add activity feature to timeout step

ikedam:
[JENKINS-26521](https://issues.jenkins-ci.org/browse/JENKINS-26521)
{quote}
Some of our test-suite items can occasionally deadlock, causing a build to never finish. The Build Timeout Plugin supports failing a job when nothing happens (i.e. no log output) for a sufficient length of time. It would be good to get this behaviour in Workflow.
{quote}

This change adds "activity" option, which makes timeout steps to watch not timeout from starts of blocks, but from last log outputs.
This works just like [`NoActivityTimeOutStrategy` of Build-Timeout plugin](https://github.com/jenkinsci/build-timeout-plugin/blob/master/src/main/java/hudson/plugins/build_timeout/impl/NoActivityTimeOutStrategy.java)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.5</version>
+        <version>2.14</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
@@ -38,9 +38,16 @@ public class TimeoutStepTest {
         TimeoutStep s1 = new TimeoutStep(3);
         s1.setUnit(TimeUnit.HOURS);
         TimeoutStep s2 = new StepConfigTester(r).configRoundTrip(s1);
-        // assertEqualDataBoundBeans does not currently work on @DataBoundSetter:
-        assertEquals(3, s2.getTime());
-        assertEquals(TimeUnit.HOURS, s2.getUnit());
+        r.assertEqualDataBoundBeans(s1, s2);
+    }
+
+    @Test public void configRoundTripWithActivity() throws Exception {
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.5</version>
+        <version>2.14</version>
```
ikedam: To use `assertEqualDataBoundBeans` supporting `DataBoundSetter` setters in jenkins-test-harness 2.11.
(Not support for `DataBoundSetter` members yet...)

src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -38,9 +38,16 @@
         TimeoutStep s1 = new TimeoutStep(3);
         s1.setUnit(TimeUnit.HOURS);
         TimeoutStep s2 = new StepConfigTester(r).configRoundTrip(s1);
-        // assertEqualDataBoundBeans does not currently work on @DataBoundSetter:
-        assertEquals(3, s2.getTime());
-        assertEquals(TimeUnit.HOURS, s2.getUnit());
+        r.assertEqualDataBoundBeans(s1, s2);
```
ikedam: `assertEqualDataBoundBeans` now supports `DataBoundSetter` setters (not yet `DataBoundSetter` members).

jglick:
Have not yet gotten a chance to review this & #19 in detail but it occurs to me that this is probably the time to move this step into the `build-timeout` plugin, implementing `@Symbol`s there and adopting its strategies (which would need to implement Pipeline-specific logic). The main thing I am unsure of is how to make sure that the `timeout` step as implemented here is actually superseded by a step of the same name in a different plugin.

jglick:
I need to get back to these…

ikedam:
The base code is old and this request is outdated.

jglick:
The PR was basically good, so I hope this can be resurrected soon.

roni-frantchi:
> ... I hope this can be resurrected soon.

@jglick what needs to be done for that to happen..?

PR #19, opened 2016-09-17
[JENKINS-27396] Introduce elastic and id to timeout step

ikedam:
[JENKINS-27396](https://issues.jenkins-ci.org/browse/JENKINS-27396)
`elastic` decides timeout limit from the last execution duration of the block in the last successful build.
`id` is the identifier of the block to get the last execution duration.

For example:

```
node {
    timeout(id: 'main', time: 5, unit: 'MINUTES', elastic: 1.5) {
        sh './build.sh`
    }
}
```

means
- This block is identified as "main"
- Timeout when 150% of the last duration is passed.
- Use 5 minutes as the timeout limit if there's no previous builds.

This may require additional changes for these issues. I'd like to hear your opinions:
- `ElasticTimeOutStrategy` in build-timeout plugin provides more features:
  - Features
    - Use mean of several builds to calcurate timeout limit. The number can be specified.
    - Use the default value if the calcurated value is too small.
    - Specify threshold of build results to use.
  - Supporting these features might result `timeout` step have many parameters and have a complicated implementaion.
  - It might be better a) extract "timeout calcurtor" and make it an extension point. b) extract `timeout` step another plugin. (like email-ext-plugin replaces mailer-plugin)
- `timeout` step get noisy with this change.
  - It logs timeout limit when starting.
  - It logs duration when finishing.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
@@ -2,8 +2,22 @@ package org.jenkinsci.plugins.workflow.steps;
 
 import com.google.inject.Inject;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.EnvVars;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+
+import java.io.IOException;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+
+import javax.annotation.CheckForNull;
+
+import org.apache.commons.lang.StringUtils;
+import org.jenkinsci.plugins.workflow.actions.TimeoutInfoAction;
+import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
+import org.jenkinsci.plugins.workflow.graph.FlowGraphWalker;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
 
 import jenkins.model.CauseOfInterruption;
 import jenkins.util.Timer;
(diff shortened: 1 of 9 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -62,13 +174,34 @@ public void stop(Throwable cause) throws Exception {
...
+    }
+
     private class Callback extends BodyExecutionCallback.TailCall {
 
         @Override protected void finished(StepContext context) throws Exception {
-            if (killer!=null) {
-                killer.cancel(true);
-                killer = null;
-            }
+            TimeoutStepExecution.this.finished();
```
ikedam: Extracted codes executed when finished to `TimeoutStepExecution`.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -88,5 +221,21 @@ public String getShortDescription() {
         }
     }
 
+    private static final class EnvironmentExpanderImpl extends EnvironmentExpander {
```
ikedam: I believe I can define a variable available in the timeout block, but I couldn't find how to do that.

src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepRunTest.java
```diff
@@ -151,6 +154,164 @@ public void evaluate() throws Throwable {
...
+    @Test
+    public void elastic() throws Exception {
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(""
+                        + "node {\n"
+                        + "  timeout(id: 'main', time: 20, unit: 'SECONDS', elastic: 1.0) {\n"
+                        + "    assert('20000' == \"${env.timeout}\");\n"
```
ikedam: I found the build stacks when `assert` failed.
I'll see why it happens.

ikedam:
This change completely conflicts with #18.
I'll rebase another afther one is merged.

src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepRunTest.java
ikedam: Looks [JENKINS-34488](https://issues.jenkins-ci.org/browse/JENKINS-34488)

jglick:
> This change completely conflicts with #18.

You can always base one PR on another, if you are feeling optimistic.

alecharp:
@ikedam do you know when you will be able to finish this PR?

ikedam:
@alecharp No, I don't have time to work at all...
Anyway, the base of this change is old and it's time to abondon this request.
I believe it's better to provide plugins a mechanism to replace basic steps as @jglick pointed in #18.

jglick:
I am less sure about that at this point. `TimeoutStepExecution` has evolved considerably, in response to bug reports and subtle changes to Pipeline internals. Making that kind of functionality live in the old `build-timeout` plugin seems less realistic. I think it would make more sense to try to merge this with the current code base—the conflicts are probably going to be superficial.

I do think it would make sense to have a timeout strategy be its own `Describable & ExtensionPoint` with a `@Symbol` so you could say something like

```groovy
timeout(fixed(time: 1, unit: 'HOURS')) {…}
// same as current syntax
```

or

```groovy
timeout(elastic(1.5)) {…}
```

or

```groovy
timeout(inactivity(time: 5, unit: 'MINUTES')) {…}
```

Sorry I have been so unresponsive here.

PR #20, opened 2016-09-19
First work on a matrix step using GroovyStep

abayer:
This is a first proof-of-concept of using https://github.com/jenkinsci/workflow-cps-plugin/pull/59. I need to add tests, etc, and we may not actually want to use this approach in the real world anyway, but...proof of concept!

cc @reviewbybees esp @jglick @kohsuke

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,18 +90,17 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.4</version>
+            <version>1.5</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.10</version>
-            <scope>test</scope>
+            <version>2.18-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.10</version>
+            <version>2.18-SNAPSHOT</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/MatrixStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/MatrixStep.java
@@ -1,0 +1,84 @@
+package org.jenkinsci.plugins.workflow.steps;
+
+import com.google.common.collect.Lists;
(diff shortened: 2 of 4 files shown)


abayer:
Added a test. This is still gonna fail on jenkins.ci.cloudbees.com due to the SNAPSHOT dependency on the other PR.

pom.xml
```diff
@@ -90,18 +90,17 @@
...
             <artifactId>structs</artifactId>
-            <version>1.4</version>
+            <version>1.5</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.10</version>
-            <scope>test</scope>
+            <version>2.18-SNAPSHOT</version>
```
jglick: :-1:, do not want this dep here. You can put it in Pipeline Utility Steps, wherever, but I want this plugin kept free of heavy deps.

src/main/java/org/jenkinsci/plugins/workflow/steps/MatrixStep.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class MatrixStep extends GroovyStep {
+    private Map<String,List<Object>> axes;
+
+    @DataBoundConstructor
+    public MatrixStep(@Nonnull Map<String,List<Object>> axes) {
+        this.axes = axes;
+    }
+
+    public Map<String,List<Object>> getAxes() {
```
jglick: `Object`?

jglick: Without a `config.jelly` this is pointless—you could just create a library, and more easily/expressively to boot.

pom.xml
abayer: Fair enough. This is just a POC anyway.

src/main/java/org/jenkinsci/plugins/workflow/steps/MatrixStep.java
abayer: Yes, again, POC. =) The `Object` is 'cos I wanted to make sure `String`s, `int`s and `boolean`s all worked.

jglick: But these _cannot_ all work in an actual `Step`. You can only use `String`s, since that is all databinding offers.

abayer: *facepalm*

Duh.

abayer:
Closing - dead.

PR #22, opened 2016-09-23
Add test case to reproduce JENKINS-34637.

orrc:
https://issues.jenkins-ci.org/browse/JENKINS-34637


jglick:
No clue offhand, would need to debug it.

When offering a test case to reproduce a bug (which is great BTW!), add

```java
@Ignore("TODO fails: blah blah blah happens instead")
```

so that your PR can be merged as is.

orrc:
Doh, indeed.

PR #23, opened 2016-10-06
[JENKINS-38640] [JENKINS-36914] Warn users to avoid stash/unstash of large files

jglick:
Cf. https://github.com/jenkinsci/workflow-api-plugin/pull/19. Suggested by jfemia on #jenkins.

@reviewbybees


amuniz:
Maybe https://github.com/jenkinsci/workflow-api-plugin/pull/19 is making this not that wrong.

abayer:
:bee:

PR #24, opened 2016-10-18
[JENKINS-34637] Test for timeout bug

jglick:
[JENKINS-34637](https://issues.jenkins-ci.org/browse/JENKINS-34637)

Downstream of https://github.com/jenkinsci/workflow-cps-plugin/pull/76, the actual fix. Incorporates #22 (thanks!).

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
@@ -2,6 +2,8 @@ package org.jenkinsci.plugins.workflow.steps;
 
 import com.google.inject.Inject;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -45,17 +48,24 @@ public void onResume() {
...
-            }, end - now, TimeUnit.MILLISECONDS);
+            }, delay, TimeUnit.MILLISECONDS);
         } else {
-            body.cancel(new ExceededTimeout());
+            cancel();
         }
     }
 
+    private void cancel() {
+        listener.getLogger().println("Cancelling nested steps due to timeout");
```
jglick: Only effective after https://github.com/jenkinsci/workflow-support-plugin/pull/19.

PR #26, opened 2016-10-24
[JENKINS-39072] timeout improvements

jglick:
[JENKINS-39072](https://issues.jenkins-ci.org/browse/JENKINS-39072)

@reviewbybees esp. @kohsuke, & @ikedam


abayer:
:bee:

ikedam:
If the build continues even after you click the "[x]" button of the build,
a message "Click here to forcibly terminate running steps" will be displayed in the console log.
And the build continues even after clicking that message, a message
"Click here to forcibly kill entire build".

Though I'm not sure who displays and handles those messages, the feature added this request works just like that.
Can't we reuse that logic?

jglick:
> I'm not sure who displays and handles those messages

`WorkflowRun`.

> Can't we reuse that logic?

In fact the `forcible` mode here *does* simulate the logic of `doTerm`, more or less. (A bit different since we are only considered steps running inside a particular block, rather than all steps in the program.)

Copying the `doKill` logic here seems too hazardous—this is an extreme measure, only necessary when there is a serious bug in one or more step implementations and/or the Groovy engine; in such cases I would rather keep the build “alive” to the extent that it can produce a virtual thread dump, etc. (There is a filed RFE to capture a virtual thread dump to the build log prior to `kill`ing.) Anyway the main reason for a `timeout` is to release executors and that sort of thing, which `kill`ing cannot do reliably since it bypasses normal cleanup mechanisms.

PR #28, opened 2016-11-30
Removing uses of Guice

jglick:
Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/15.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStep.java
@@ -24,26 +24,28 @@
 
 package org.jenkinsci.plugins.workflow.steps;
 
+import com.google.common.collect.ImmutableSet;
 import hudson.AbortException;
 import hudson.Extension;
 import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+import java.util.Set;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 /**
  * Runs a block.
  * If it fails, marks the build as failed, but continues execution.
  */
-public final class CatchErrorStep extends AbstractStepImpl {
+public final class CatchErrorStep extends Step {
 
     @DataBoundConstructor public CatchErrorStep() {}
 
-    @Extension public static final class DescriptorImpl extends AbstractStepDescriptorImpl {
+    @Override public StepExecution start(StepContext context) throws Exception {
+        return new Execution(context);
+    }
 
-        public DescriptorImpl() {
-            super(Execution.class);
-        }
+    @Extension public static final class DescriptorImpl extends StepDescriptor {
 
         @Override public String getFunctionName() {
             return "catchError";
@@ -61,13 +63,17 @@ public final class CatchErrorStep extends AbstractStepImpl {
             return true;
         }
 
+        @Override public Set<? extends Class<?>> getRequiredContext() {
+            return ImmutableSet.of(Run.class, TaskListener.class);
+        }
+
     }
 
-    public static final class Execution extends AbstractStepExecutionImpl {
+    public static final class Execution extends StepExecution {
 
-        /** Placeholder for getRequiredContext. */
-        @StepContextParameter private transient Run<?,?> run;
-        @StepContextParameter private transient TaskListener listener;
+        Execution(StepContext context) {
+            super(context);
+        }
 
         @Override public boolean start() throws Exception {
             StepContext context = getContext();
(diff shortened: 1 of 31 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStep.java
```diff
@@ -61,13 +63,17 @@ public DescriptorImpl() {
...
+        }
+
     }
 
-    public static final class Execution extends AbstractStepExecutionImpl {
+    public static final class Execution extends StepExecution {
 
-        /** Placeholder for getRequiredContext. */
-        @StepContextParameter private transient Run<?,?> run;
-        @StepContextParameter private transient TaskListener listener;
```
jglick: One of the evil tricks that is no longer necessary.

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
```diff
@@ -48,42 +49,44 @@
...
+    }
+
+    private static final class Execution extends SynchronousNonBlockingStepExecution<Void> {
 
-        @Inject private transient CoreStep step;
-        @StepContextParameter private transient Run<?,?> run;
-        @StepContextParameter private transient FilePath workspace;
-        @StepContextParameter private transient Launcher launcher;
-        @StepContextParameter private transient TaskListener listener;
+        private transient final SimpleBuildStep delegate;
```
jglick: Have the option of simplifying `StepExecution` state, especially in cases where the `Step` had only one field.

src/main/java/org/jenkinsci/plugins/workflow/steps/SleepStep.java
```diff
@@ -57,15 +62,23 @@ public TimeUnit getUnit() {
...
+        return new Execution(this, context);
+    }
+
+    public static final class Execution extends StepExecution {
 
         private static final long serialVersionUID = 1L;
 
-        @Inject(optional=true) private transient SleepStep step;
-        @StepContextParameter private transient TaskListener listener;
+        private transient final SleepStep step;
```
jglick: Alternately could have kept just the result of `step.unit.toMillis(step.time)`.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -51,10 +53,18 @@ public boolean start() throws Exception {
...
-        super.onResume();
         setupTimer(System.currentTimeMillis());
     }
 
+    private TaskListener listener() {
+        try {
+            return getContext().get(TaskListener.class);
+        } catch (Exception x) {
+            LOGGER.log(Level.WARNING, null, x);
+            return TaskListener.NULL;
```
jglick: When using injection, there were two ways `StepContext.get` errors were handled: `AbstractStepImpl.start` would throw up the exception, which we could still do in a `StepExecution` constructor; and `AbstractStepExecutionImpl.onResume` would send the exception to `StepContext.onFailure` (but presumably leave the field null, without preventing the rest of an `onResume` override from running—perhaps the root cause of JENKINS-34021 and the still-open [JENKINS-37486](https://issues.jenkins-ci.org/browse/JENKINS-37486)).

This idiom is more verbose but clearer—the step gets to decide what to do.

Alternately, we could change `onResume` to throw `Exception`. (Would still be binary-compatible.) `FlowExecutionList.ItemListenerImpl.onLoaded` would then need to catch exceptions and call `StepContext.onFailure`. Like `AbstractStepExecutionImpl`, this would rely on `transient` fields holding the contextual object; the improvement would be that a failure in `StepContext.get` would stop anything _else_ in `onResume` (such as `setupTimer` here) from running.

stephenc:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
jglick: Filed [JENKINS-40161](https://issues.jenkins-ci.org/browse/JENKINS-40161) to track the latter proposal.

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
```diff
@@ -50,22 +52,32 @@
...
     public List<String> getOverrides() {
         return overrides;
     }
+
+    @Override public StepExecution start(StepContext context) throws Exception {
+        return new Execution(overrides, context);
+    }
     
-    public static class Execution extends AbstractStepExecutionImpl {
+    public static class Execution extends StepExecution {
```
jglick: Likely to cause [JENKINS-40909](https://issues.jenkins-ci.org/browse/JENKINS-40909).

PR #33, opened 2017-01-15
Introduce a requireSteps step

batmat:
Some followup of https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/31 which made me think it would be nice to be more specific. Also, it's sometimes a headache for users to be sure where steps should be coming from.

Sample usage:

```groovy
requireSteps steps:['pwd','findFiles']
```

NOTE: this checks only on functions and symbols, not on global variables. @jglick I am not sure how to do that, apart from adding a dependency against groovy-cps which I suspect is probably not the way to go right?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/RequireStepsStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/RequireStepsStep.java
@@ -1,0 +1,149 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Baptiste Mathus.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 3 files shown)


jglick:
> it's sometimes a headache for users to be sure where steps should be coming from

Which is why we need **Pipeline Syntax** support, which is what I was talking about in #31. And that is not possible until @svanoort finishes https://github.com/jenkinsci/workflow-api-plugin/pull/26.

svanoort:
@jglick I'm rather unclear why the StepAction is a precursor for pipeline syntax support?

I am only implementing JENKINS-37324, not JENKINS-31582 here -- perhaps there's a mapping from the Descriptor that I'm not aware of?

HRMPW:
As I said in PR-31, I think the best user experience for this would be to build this into Declarative to be automatic. `steps` are explicit in Declarative except for anything in a `script` block. Making a user type in a list of `steps` they want to use when we already know the list of `steps` that are being used seems like an unnecessary ask of the Pipeline builder. 

If we build this for Scripted and require the user to list the `steps` included in the Pipeline then that is fine but for Declarative we should automatically fill the list of `steps` used for the user.

batmat:
Closing in favor of https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/31 as per apparent trend above. Cleaning up my open PR list a bit.

PR #34, opened 2017-01-25
[JENKINS-41276] Do not retry if user aborted the build

rsandell:
See [JENKINS-41276](https://issues.jenkins-ci.org/browse/JENKINS-41276)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
@@ -67,13 +67,11 @@ public class RetryStepExecution extends AbstractStepExecutionImpl {
             try {
                 Run run = context.get(Run.class);
                 if (run != null && t instanceof FlowInterruptedException) {
-                    InterruptedBuildAction action = run.getAction(InterruptedBuildAction.class);
-                    if (action != null) {
-                        for (CauseOfInterruption cause : action.getCauses()) {
-                            if (cause instanceof CauseOfInterruption.UserInterruption) {
-                                context.onFailure(t);
-                                return;
-                            }
+                    FlowInterruptedException fie = (FlowInterruptedException) t;
+                    for (CauseOfInterruption cause : fie.getCauses()) {
+                        if (cause instanceof CauseOfInterruption.UserInterruption) {
+                            context.onFailure(t);
+                            return;
                         }
                     }
                 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
@@ -2,7 +2,11 @@ package org.jenkinsci.plugins.workflow.steps;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.AbortException;
+import hudson.model.Result;
+import hudson.model.Run;
 import hudson.model.TaskListener;
+import jenkins.model.CauseOfInterruption;
+import jenkins.model.InterruptedBuildAction;
 
 /**
  * @author Kohsuke Kawaguchi
@@ -61,6 +65,18 @@ public class RetryStepExecution extends AbstractStepExecutionImpl {
         @Override
         public void onFailure(StepContext context, Throwable t) {
             try {
+                Run run = context.get(Run.class);
+                if (run != null && t instanceof FlowInterruptedException) {
+                    InterruptedBuildAction action = run.getAction(InterruptedBuildAction.class);
+                    if (action != null) {
+                        for (CauseOfInterruption cause : action.getCauses()) {
+                            if (cause instanceof CauseOfInterruption.UserInterruption) {
+                                context.onFailure(t);
+                                return;
+                            }
+                        }
+                    }
+                }
                 left--;
                 if (left>0) {
                     TaskListener l = context.get(TaskListener.class);
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
@@ -2,7 +2,6 @@ package org.jenkinsci.plugins.workflow.steps;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.AbortException;
-import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import jenkins.model.CauseOfInterruption;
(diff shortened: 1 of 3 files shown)


stephenc:
:bee:

rsandell:
@reviewbybees done

src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
```diff
@@ -61,6 +65,18 @@ public void onSuccess(StepContext context, Object result) {
         @Override
         public void onFailure(StepContext context, Throwable t) {
             try {
+                Run run = context.get(Run.class);
+                if (run != null && t instanceof FlowInterruptedException) {
+                    InterruptedBuildAction action = run.getAction(InterruptedBuildAction.class);
+                    if (action != null) {
+                        for (CauseOfInterruption cause : action.getCauses()) {
```
jglick: No, use `FlowInterruptedException.getCauses()`, then you do not need to refer to `Run` or `InterruptedBuildAction` at all.

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
```diff
@@ -0,0 +1,47 @@
...
+ * Tests {@link RetryStep}.
+ */
+public class RetryStepTest {
+
+    @ClassRule
+    public static BuildWatcher buildWatcher = new BuildWatcher();
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @Test(timeout = 50000)
```
jglick: Delete, all `JenkinsRule` tests have an implicit timeout anyway.

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import java.io.IOException;
+
+import static org.junit.Assert.*;
+
+/**
+ * Tests {@link RetryStep}.
```
jglick: BTW `WorkflowJobNonRestartingTest.testRetry` should really be moved here from `workflow-cps`.

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
```diff
@@ -0,0 +1,47 @@
...
+    public static BuildWatcher buildWatcher = new BuildWatcher();
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @Test(timeout = 50000)
+    @Issue("JENKINS-41276")
+    public void abortShouldNotRetry() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(
+                "int count = 0; retry(3) { echo 'trying '+(count++); semaphore 'start'; sleep 10; echo 'NotHere' }", false));
```
jglick: Please use `true` for the `sandbox` argument unless you have a compelling reason to test trusted code specifically.

jglick: The `sleep` is gratuitous here. Delete it to make the test faster.

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
```diff
@@ -0,0 +1,47 @@
...
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("start/1", b);
+        b.doStop();
+        b = r.assertBuildStatus(Result.ABORTED, r.waitForCompletion(b));
+        r.assertLogContains("trying 0", b);
+        r.assertLogContains("Aborted by anonymous", b);
+        r.assertLogNotContains("trying 1", b);
+        r.assertLogNotContains("trying 2", b);
+        r.assertLogNotContains("NotHere", b);
```
jglick: Gratuitous check.

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
rsandell: Yes, I was originally intending to let the script abort itself, but went another route and the false sandbox was left.

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
rsandell: Well that timeout was too long for a reasonable test performance

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
rsandell: Crossing the t and dotting the i.

src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
rsandell: I tried that in 3c2bba9 but the exception didn't contain any causes to check. See https://jenkins.ci.cloudbees.com/job/plugins/job/workflow-basic-steps-plugin/org.jenkins-ci.plugins.workflow$workflow-basic-steps/187/

So reverted that change in 3c9dfec

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
jglick: When the test is passing, or failing? If failing, we do not care. If passing, then the test needs to be fixed to pass quickly.

src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
jglick: I suspect [this](https://github.com/jenkinsci/workflow-job-plugin/blob/139ac7e5da1e03ec16be698a698a3574aa748722/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L273) needs to call `getCausesOfInterruption()`. While it already does call `recordCauseOfInterruption`, which records information in the `InterruptedBuildAction`, this does not allow the causes to be inspected by catching code, which is no good.

The call to `recordCauseOfInterruption` could also be removed, since `FlowInterruptedException.handle` would already add it. On the other hand, for a _caught_ exception we do currently record interruption cause, which is possibly useful especially if several people attempt to interrupt it at different times. So the question is whether we can record it for both APIs without printing duplicate messages to the log. Needs to be verified in `workflow-job`.

BTW `Executor` only records a `UserInterruption` if the interrupt is sent by a real user, not `SYSTEM`, so I am curious how your test could pass given that it does not `impersonate` anyone.

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
jglick: Worse, the test _fails_ when the `timeout` is removed. I think because of a race condition: the call to `recordCauseOfInterruption` from `WorkflowRun` happens asynchronously. So the `InterruptedBuildAction` may or may not be present at the time the `onFailure` handler is called, according to timing. So the current patch is doomed.

PR #39, opened 2017-03-24
[JENKINS-26137] Picking up possible fix for integration testing

jglick:
Attempting to fix [this error](https://ci.jenkins.io/job/Plugins/job/workflow-basic-steps-plugin/job/PR-37/3/testReport/org.jenkinsci.plugins.workflow.steps/TimeoutStepTest/serialForm_2/):

```
…
Resuming build at … after Jenkins restart
No need to sleep any longer
[Pipeline] }
[Pipeline] // timeout
[Pipeline] End of Pipeline
java.lang.IllegalStateException: JENKINS-26137: Jenkins is shutting down
 at org.jenkinsci.plugins.workflow.support.pickles.serialization.RiverWriter.<init>(RiverWriter.java:92)
 at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.saveProgram(CpsThreadGroup.java:431)
 at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.saveProgram(CpsThreadGroup.java:412)
 at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.run(CpsThreadGroup.java:357)
 at …
```

Downstream of https://github.com/jenkinsci/workflow-support-plugin/pull/35 & https://github.com/jenkinsci/workflow-cps-plugin/pull/117.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.21</version>
+        <version>2.25</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -64,6 +64,8 @@
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
         <workflow-step-api-plugin.version>2.9</workflow-step-api-plugin.version>
+        <workflow-cps-plugin.version>2.30-20170324.022019-1</workflow-cps-plugin.version> <!-- TODO https://github.com/jenkinsci/workflow-cps-plugin/pull/117 -->
+        <workflow-support-plugin.version>2.14-20170324.021801-1</workflow-support-plugin.version> <!-- TODO https://github.com/jenkinsci/workflow-support-plugin/pull/35 -->
     </properties>
     <dependencies>
         <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -86,23 +88,29 @@
...
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.5</version>
+            <version>1.6</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.22</version>
+            <version>${workflow-cps-plugin.version}</version>
```
svanoort: 🐜  Please stop doing this --  if the version is only used twice *right above each other*, it makes the code less readable because now I have to jump up to property declarations to see the actual version used.

pom.xml
```diff
@@ -86,23 +88,29 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.8</version>
+            <version>2.11</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jenkins-ci.plugins</groupId>
+                    <artifactId>scm-api</artifactId>
```
oleg-nenashev: Would be better to avoid it if possible. Does it require a version bump?

pom.xml
```diff
@@ -121,13 +129,13 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
-            <version>2.10</version>
+            <version>${workflow-support-plugin.version}</version>
```
svanoort: 🐜 Again, this actually makes it MORE work to read and interact with the pom because now the version I want to change is separated from the usage.  

This would ONLY make sense if you had 3+ usages, or the usages were not *right next* to each other.

pom.xml
jglick: I have had numerous cases where I forgot to update one or the other version, so I am instituting a policy going forward of always using a property whenever a plugin and its tests are both used as dependencies, regardless of visual distance in the POM.

pom.xml
jglick: Avoid what? I am asking to pick up the `scm-api` dependency from some other dependency trail, not this one.

pom.xml
svanoort: Is there a way you could approach this that would *not* require jumping back and forth, perhaps by tying the test version to the main version?

jglick: Not that I am aware of.

jglick:
@reviewbybees done

(but deferring merge since I would like to get https://github.com/jenkinsci/workflow-cps-plugin/pull/120 in the next upstream release)

PR #40, opened 2017-04-01
Add global variable for accessing the proxy settings in pipelines

coder-hugo:
I haven't found any existing solution for this and I also haven't got an answer for [my question in the jenkins-users group](https://groups.google.com/forum/#!msg/jenkinsci-users/MFsgYtWUOZw/uS5zgTmDCgAJ;context-place=forum/jenkinsci-users). Therefore I just implemented a global variable that provides access to the Jenkins proxy settings in pipelines.
I think this plugin is a good place for such a variable. But if you are thinking differently I can move this also to a separate plugin.


pom.xml
```diff
@@ -97,7 +97,6 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
             <version>2.22</version>
-            <scope>test</scope>
```
jglick: Do not want this dep here.

coder-hugo:
I just changed my PR according to your suggestion. There is now a step (`withHttpProxyEnv`) that provides the global proxy settings as the following environment variables:
* `http_proxy`/`HTTP_PROXY`
* `https_proxy`/`HTTPS_PROXY`
* `no_proxy`/`NO_PROXY`

I add both the lower case and the upper case format as there are commands that like just the lower case format and others that just like the upper case one.

Sh-Ho:
Hello, 
Is this fix about to be merged? is there another solution for this problem?
Thanks!

PR #41, opened 2017-04-05
[JENKINS-29037] Allow for customization of the wait for condition step

mmitche:
Fixes an issue we were seeing where the step by default tends to run the contained closure very often initially.  When running lots of pipelines in parallel, this could cause problems if we had something like an http request within the body.  Allows for additional parameters to set the minimum recurrence period, the maximum recurrence period, and the units.


jakauppila:
This would be great to have. We're seeing the same problem with the step as-is, we have some HTTP calls to check the status of another process that takes 60 min+, so we want to severely limit when and how often we're checking that.

src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
```diff
@@ -80,7 +116,6 @@
         }
 
         @Override public void onResume() {
-            recurrencePeriod = MIN_RECURRENCE_PERIOD;
```
jglick: You need to restore `currentRecurrencePeriod` to the minimum here; why was this deleted?

src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
```diff
@@ -56,16 +89,19 @@
          */
         private final String id = UUID.randomUUID().toString();
         private static final float RECURRENCE_PERIOD_BACKOFF = 1.2f;
-        static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
-        static final long MAX_RECURRENCE_PERIOD = 15000; // ¼min
-        long recurrencePeriod = MIN_RECURRENCE_PERIOD;
-
-        Execution(StepContext context) {
+        long currentRecurrencePeriod;
```
jglick: The rename of the `recurrencePeriod` makes the rest of the diff harder to follow.

src/main/resources/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep/config.jelly
```diff
@@ -24,4 +24,15 @@ THE SOFTWARE.
 -->
 
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core"/>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="minRecurrencePeriod" title="Minimum Recurrence Period">
```
jglick: use `${%Some localizable text}` syntax

jglick:
I think this purports to solve [JENKINS-29037](https://issues.jenkins-ci.org/browse/JENKINS-29037); marking accordingly.

i386:
@mmitche hey there - are you still working on this one? is there any way we can help?

mmitche:
@i386 I'll fix this up today.  I think I have the changes locally.

mmitche:
Feedback addressed

i386:
what say you @jglick ?

i386:
@mmitche thanks mate!

src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
```diff
@@ -56,16 +89,23 @@
...
         private final String id = UUID.randomUUID().toString();
         private static final float RECURRENCE_PERIOD_BACKOFF = 1.2f;
-        static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
-        static final long MAX_RECURRENCE_PERIOD = 15000; // ¼min
-        long recurrencePeriod = MIN_RECURRENCE_PERIOD;
-
-        Execution(StepContext context) {
+        long recurrencePeriod;
+        long maxRecurrencePeriod;
+        long minRecurrencePeriod;
```
jglick: When you are introducing non`transient` fields to a `Serializable` class you must define reasonable defaults for them in a `readResolve` method. In this case it means that a `waitUntil` block started prior to plugin upgrade will see zero values for `maxRecurrencePeriod` and `minRecurrencePeriod`, probably causing mayhem.

src/main/resources/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep/config.jelly
```diff
@@ -24,4 +24,15 @@ THE SOFTWARE.
...
-<j:jelly xmlns:j="jelly:core"/>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="minRecurrencePeriod" title="${%Minimum Recurrence Period}">
+        <f:number clazz="positive-number"/>
+    </f:entry>
+    <f:entry field="maxRecurrencePeriod" title="${%Maximum Recurrence Period}">
+        <f:number clazz="positive-number"/>
+    </f:entry>
+    <f:entry field="unit" title="Unit">
+        <f:select/>
```
jglick: Should set a `default`.

src/main/resources/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep/config.jelly
```diff
@@ -24,4 +24,15 @@ THE SOFTWARE.
 -->
 
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core"/>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="minRecurrencePeriod" title="${%Minimum Recurrence Period}">
+        <f:number clazz="positive-number"/>
+    </f:entry>
+    <f:entry field="maxRecurrencePeriod" title="${%Maximum Recurrence Period}">
+        <f:number clazz="positive-number"/>
```
jglick: You need a `default` here, which would most easily be a constant in `DescriptorImpl`, moved there from the `Step` itself. [Documentation](https://github.com/jenkinsci/pipeline-plugin/blob/master/DEVGUIDE.md#handling-default-values)

mmitche:
@jglick Responded to feedback.  I only added the min/max to the readResolve since recurrencePeriod will be set on resume.

mmitche:
@jglick Anything else here?

mmitche:
@abayer Thanks

mmitche:
Anything else or can this be merged?

szpak:
@abayer Have you been able to merge it? It seems still to be "in review".

src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
```diff
@@ -56,16 +91,29 @@
...
+        
+        private Object readResolve() {
+            minRecurrencePeriod = DescriptorImpl.DEFAULT_MIN_RECURRENCE_PERIOD;
+            maxRecurrencePeriod = DescriptorImpl.DEFAULT_MAX_RECURRENCE_PERIOD;
+            return this;
         }
 
         @Override public boolean start() throws Exception {
             body = getContext().newBodyInvoker().withCallback(new Callback(id)).start();
+            this.recurrencePeriod = step.unit.toMillis(step.getMinRecurrencePeriod());
```
dwnusbaum: Would `step.unit` be null here for running builds if we upgraded the plugin and then restarted  Jenkins?

PR #42, opened 2017-04-23
[JENKINS-37327] allow empty stash.

iwarapter:
Depends on https://github.com/jenkinsci/workflow-api-plugin/pull/35


src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/config.jelly
```diff
@@ -37,4 +37,7 @@ THE SOFTWARE.
     <f:entry field="useDefaultExcludes" title="Use Default Ant Excludes">
         <f:checkbox default="true" />
     </f:entry>
+    <f:entry field="allowEmpty" title="Allow empty stash">
+        <f:checkbox default="false" />
```
jglick: BTW `default="false"` is superfluous.

iwarapter:
@jglick thanks for merging this, did you want a PR to fix the superfluous `default="false"`?

jglick:
Need not bother I think.

PR #43, opened 2017-05-16
Update stash step help to suggest size guidelines

svanoort:
Like it says on the tin.


svanoort:
@reviewbybees

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
```diff
@@ -4,4 +4,7 @@
     <strong>Note</strong> that the <code>stash</code> and <code>unstash</code> steps are designed for use with small files.
     For large data transfers, use the External Workspace Manager plugin,
     or use an external repository manager such as Nexus or Artifactory.
+    This is because stashed files are archived in a compressed TAR, and with large files this demands considerable on-master
+    resources, particularly CPU time.
```
jglick: Also agent → master network bandwidth, though I suppose the critical thing for most users is going to be CPU usage on the master.

svanoort:
@reviewbybees done

PR #46, opened 2017-06-18
Mention the default units for the timeout step.

mostynb:
src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStep.java uses
'MINUTES' as the default units.

Also, improved the language slightly.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/steps/TimeoutStep/help.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/steps/TimeoutStep/help.html
@@ -1,5 +1,6 @@
 <div>
-    Executes the code inside the block with a determined time out limit.
-    If the time limit is reached, an exception is thrown, which leads in aborting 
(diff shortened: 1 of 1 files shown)


mostynb:
I guess this is a duplicate of:
https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/13

mostynb:
It is not clear to me what changes you're requesting...

jglick:
Probably easiest for me to file my own PR for it.

mostynb:
If you provide some more details (maybe we could add a contributing.md?) then it might be easier for other people to help you out more.

abayer:
so what @jglick was saying on #13 is that there should be separate `help` html files for the individual parameters - i.e., `help-unit.html` and `help-time.html`. That'll result in the snippet generator showing the help button alongside the parameters, rather than just for the entire step.

dwnusbaum:
See https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/114.

PR #47, opened 2017-06-30
[JENKINS-45101] argumentsToString implementations

jglick:
[JENKINS-45101](https://issues.jenkins-ci.org/browse/JENKINS-45101)

Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/27.

Does not cover every case. Most notably,

```groovy
junit testResults: '*.xml', keepLongStdio: true
```

is not supported, because this would require changes in `workflow-api`. (`argumentsToString` is currently given an unresolved map, which `workflow-basic-steps` _could_ resolve in the same manner as `ArgumentsAction.getResolvedArguments`, but then it would be duplicating code which is also due for a rewrite as described in comments there.) Good enough for now.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,7 @@
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
-        <workflow-step-api-plugin.version>2.11</workflow-step-api-plugin.version>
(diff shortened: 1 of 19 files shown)


jglick:
@reviewbybees done

PR #48, opened 2017-08-21
[FIXED JENKINS-46179, JENKINS-46178] Clarify archive deprecation

abayer:
[JENKINS-46178](https://issues.jenkins-ci.org/browse/JENKINS-46178), [JENKINS-46179](https://issues.jenkins-ci.org/browse/JENKINS-46179)

Both in help message in snippet generator and a message echoed to the
console log when the step is actually run.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/ArtifactArchiverStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/ArtifactArchiverStepExecution.java
@@ -32,6 +32,10 @@ public class ArtifactArchiverStepExecution extends SynchronousNonBlockingStepExe
     protected Void run() throws Exception {
         FilePath ws = getContext().get(FilePath.class);
         ws.mkdirs();
+        TaskListener listener = getContext().get(TaskListener.class);
+        if (listener != null) {
+            listener.getLogger().println("The 'archive' step is deprecated and may not work properly, please use 'archiveArtifacts' instead.");
+        }
         Map<String,String> files = ws.act(new ListFiles(step.getIncludes(), step.getExcludes()));
         getContext().get(Run.class).pickArtifactManager().archive(ws, getContext().get(Launcher.class), new BuildListenerAdapter(getContext().get(TaskListener.class)), files);
         return null;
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/ArtifactArchiverStepExecution.java
```diff
@@ -32,6 +32,10 @@
     protected Void run() throws Exception {
         FilePath ws = getContext().get(FilePath.class);
         ws.mkdirs();
+        TaskListener listener = getContext().get(TaskListener.class);
+        if (listener != null) {
+            listener.getLogger().println("The 'archive' step is deprecated and may not work properly, please use 'archiveArtifacts' instead.");
```
jglick: Well it works “properly” AFAIK, it is just unnecessary when you are on 2.x…so why not just bump up the Jenkins baseline to 2.7.3 like in other plugins, at which point you can simplify a whole bunch of test cases in this plugin which had been switching modes based on `Class.hasAnnotation`?

src/test/java/org/jenkinsci/plugins/workflow/steps/ArtifactArchiverStepTest.java
```diff
@@ -41,6 +41,8 @@ public void archive() throws Exception {
         VirtualFile archivedFile = b.getArtifactManager().root().child("msg.out");
         assertTrue(archivedFile.exists());
         assertEquals("hello world", IOUtils.toString(archivedFile.open()));
+        j.assertLogContains("The 'archive' step is deprecated and may not work properly, please use 'archiveArtifacts' instead.",
```
jglick: If you are going to assert a user message, put it in `Message.properties`.

(There is a longstanding question whether strings used in build logs should be localized, since the user’s native language may not match the locale of the server, but merely creating an English key harms nothing.)

jglick:
A 2.x baseline bump could also allow `ArtifactArchiverStep.DescriptorImpl.isAdvanced` to `return true;` unconditionally.

src/main/java/org/jenkinsci/plugins/workflow/steps/ArtifactArchiverStepExecution.java
abayer: Will do.

src/test/java/org/jenkinsci/plugins/workflow/steps/ArtifactArchiverStepTest.java
abayer: Duh on me. Will do.

PR #49, opened 2017-08-29
Add an optional initialRecurrencePeriod arg to the waitUntil() step.

llamahunter:
Starting from 250msec and ramping up through two orders of magnitude for very slow operations causes a lot of log spew. Allow setting the initial recurrence period to something other than the default.  We often set it to 15 seconds, as we know up front that we'll need to wait about a minute for some async operations.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
@@ -49,7 +49,7 @@ public final class WaitForConditionStep extends Step {
 
     @DataBoundSetter
     public void setInitialRecurrencePeriod(long initialRecurrencePeriod) {
-        this.initialRecurrencePeriod = Math.min(initialRecurrencePeriod, MAX_RECURRENCE_PERIOD);
+        this.initialRecurrencePeriod = Math.max(MIN_RECURRENCE_PERIOD, Math.min(initialRecurrencePeriod, MAX_RECURRENCE_PERIOD));
     }
 
     long getInitialRecurrencePeriod() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
@@ -36,13 +36,28 @@ import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 import jenkins.util.Timer;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
 public final class WaitForConditionStep extends Step {
 
+    static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
+    static final long MAX_RECURRENCE_PERIOD = 15000; // ¼min
+
+    private long initialRecurrencePeriod = MIN_RECURRENCE_PERIOD;
+
     @DataBoundConstructor public WaitForConditionStep() {}
 
+    @DataBoundSetter
+    public void setInitialRecurrencePeriod(long initialRecurrencePeriod) {
+        this.initialRecurrencePeriod = Math.min(initialRecurrencePeriod, MAX_RECURRENCE_PERIOD);
+    }
+
+    long getInitialRecurrencePeriod() {
+        return initialRecurrencePeriod;
+    }
+
     @Override public StepExecution start(StepContext context) throws Exception {
-        return new Execution(context);
+        return new Execution(context, initialRecurrencePeriod);
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
```diff
@@ -56,12 +71,13 @@
          */
         private final String id = UUID.randomUUID().toString();
         private static final float RECURRENCE_PERIOD_BACKOFF = 1.2f;
-        static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
-        static final long MAX_RECURRENCE_PERIOD = 15000; // ¼min
-        long recurrencePeriod = MIN_RECURRENCE_PERIOD;
+        private final long initialRecurrencePeriod;
```
dwnusbaum: If we deserialize instances created before this field was added (might happen for a long running pipeline if we update the plugin and then restart?), this will be 0, which looks like it will break the logic. Maybe make it non final and set it to MIN_RECURRENCE_PERIOD by default, or add a `readResolve` method?

src/main/resources/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep/config.jelly
```diff
@@ -24,4 +24,8 @@ THE SOFTWARE.
 -->
 
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core"/>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="initialRecurrencePeriod" title="Initial Recurrence Period Milliseconds">
```
dwnusbaum: Is it worth adding a help page (`help-intialRecurrencePeriod.html`) for this field?

src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
```diff
@@ -56,12 +71,13 @@
...
+        long recurrencePeriod;
 
-        Execution(StepContext context) {
```
dwnusbaum: Any binary compatibility concerns with deleting this constructor? Doesn't look like there are any callers based on a [quick search](https://github.com/search?utf8=%E2%9C%93&q=user%3Ajenkinsci+WaitForConditionStep&type=Code), but I might not be looking in the right places.

src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
jglick: Exactly, this needs to be restored in `readResolve`, or the code reading it needs to handle 0.

src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
jglick: It is package-private, so no.

dwnusbaum: Whoops, missed that

llamahunter:
Ok, made recommended adjustments.

llamahunter:
poke... can this get merged someday?

src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java
```diff
@@ -36,13 +36,28 @@
...
+    static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
+    static final long MAX_RECURRENCE_PERIOD = 15000; // ¼min
+
+    private long initialRecurrencePeriod = MIN_RECURRENCE_PERIOD;
+
     @DataBoundConstructor public WaitForConditionStep() {}
 
+    @DataBoundSetter
+    public void setInitialRecurrencePeriod(long initialRecurrencePeriod) {
+        this.initialRecurrencePeriod = Math.min(initialRecurrencePeriod, MAX_RECURRENCE_PERIOD);
```
dwnusbaum: I think you also need to make sure this is at least `MIN_RECURRENCE_PERIOD`:
```suggestion
        this.initialRecurrencePeriod = Math.max(MIN_RECURRENCE_PERIOD, Math.min(initialRecurrencePeriod, MAX_RECURRENCE_PERIOD));
```

llamahunter:
Sure.  Didn't account for users setting initial recurrence period shorter than the minimum.  The main reason I wanted this was to be able to set the initial recurrence period *longer* than the current 250ms value.

dwnusbaum:
Yeah, my main concern was that although the UI had client-side validation for positive numbers nothing on the server side had similar validation. Thanks for the PR, I'll merge once the CI build passes!

PR #51, opened 2017-09-15
[FIXED JENKINS-40333] catchError should still add script approvals

abayer:
[JENKINS-40333](https://issues.jenkins-ci.org/browse/JENKINS-40333)

The output of the build is the same, but if we see a
RejectedAccessException, we add that to ScriptApproval as well.

cc @reviewbybees


jglick:
I am solving this more directly.

PR #52, opened 2017-10-03
[FIXED JENKINS-40912] Return list of files from stash/unstash steps

abayer:
[JENKINS-40912](https://issues.jenkins-ci.org/browse/JENKINS-40912)

downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/56

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/UnstashStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/UnstashStep.java
@@ -30,6 +30,8 @@ import hudson.Extension;
 import hudson.FilePath;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+
+import java.util.List;
 import java.util.Set;
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
@@ -58,7 +60,7 @@ public class UnstashStep extends Step {
         return new Execution(name, context);
     }
 
-    public static class Execution extends SynchronousNonBlockingStepExecution<Void> {
+    public static class Execution extends SynchronousNonBlockingStepExecution<List<String>> {
 
         private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/UnstashStep.java
```diff
@@ -70,9 +72,9 @@ public String getName() {
             this.name = name;
         }
 
-        @Override protected Void run() throws Exception {
+        @Override protected List<String> run() throws Exception {
             StashManager.unstash(getContext().get(Run.class), name, getContext().get(FilePath.class), getContext().get(TaskListener.class));
-            return null;
+            return StashManager.fileNamesForStash(getContext().get(Run.class), name);
```
svanoort: Do I understand correctly that we're extracting the files twice?  That seems... rather awkward.  Or at least something we might consider making optional and not mandatory with each run.

jglick:
I guess the question is whether we actually want this feature. The requester did not specifically ask for programmatic processing of the list of stashed files—only for some way of diagnosing a mistake in the pattern, which could perhaps be addressed in other ways: `FINE` logging, a `verbose` flag, **Pipeline Syntax** prevalidation of a proposed pattern given an example workspace, …

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
```diff
@@ -1,5 +1,6 @@
 <div>
     Saves a set of files for use later in the same build, generally on another node/workspace.
+    The step will return a list of files that are included in the stash.
```
rsandell: File names, or paths? And if paths are they relative to the current working directory or workspace or absolute?

abayer:
This sucker is moribund in favor of #54

PR #53, opened 2017-11-25
Ignore eclipse configuration files

LinuxSuRen:


LinuxSuRen:
@oleg-nenashev why this repo nobody review.

oleg-nenashev:
It depends on the plugin maintainers. Not all of them are really active.
Use @jenkinsci/code-reviewers is such case

LinuxSuRen:
So the review team is very important, when the owner no longer active or care it.

oleg-nenashev:
Yes, that's what we are trying to do, more or less successfully

LinuxSuRen:
In order to let more and more Chinese people use Jenkins, I will translate all English to Chinese step and step.

PR #54, opened 2017-12-06
[FIXED JENKINS-40912] Return a list of included files from stash/unstash

abayer:
[JENKINS-40912](https://issues.jenkins-ci.org/browse/JENKINS-40912)

Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/62

This should be more robust than the previous attempt, since it doesn't
depend on any changes to existing StashManager methods.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
@@ -30,6 +30,8 @@ import hudson.FilePath;
 import hudson.Util;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import jenkins.model.Jenkins;
@@ -98,7 +100,7 @@ public class StashStep extends Step {
         return new Execution(this, context);
     }
 
-    public static class Execution extends SynchronousNonBlockingStepExecution<Void> {
+    public static class Execution extends SynchronousNonBlockingStepExecution<List<String>> {
 
         private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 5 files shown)


jglick:
Same question as in #52 applies: is this what the user actually wanted?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
```diff
@@ -109,10 +111,10 @@ public boolean isAllowEmpty() {
             this.step = step;
         }
 
-        @Override protected Void run() throws Exception {
+        @Override protected List<String> run() throws Exception {
             StashManager.stash(getContext().get(Run.class), step.name, getContext().get(FilePath.class), getContext().get(TaskListener.class), step.includes, step.excludes,
                     step.useDefaultExcludes, step.allowEmpty);
-            return null;
+            return StashManager.stashFileList(getContext().get(Run.class), step.name);
```
jglick: This seems inefficient and silly. Just make `stash` and `unstash` return the `List<String>` as they are processed, so you do not need to process the whole tarball twice.

abayer: So the reason for this approach is that when we changed the signature for `StashManager.stash` and `StashManager.unstash` and released `workflow-api`, [users started seeing hangs with `stash`](https://issues.jenkins-ci.org/browse/JENKINS-40912?focusedCommentId=317875&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-317875) with the new `workflow-api` version. I never managed to reproduce the hang in tests, but I assumed it was due to the signature change, so this time I tried to avoid needing a signature change.

jglick:
Upstream is closed.

PR #62, opened 2018-03-07
[FIXED JENKINS-26521] Add activity flag to timeout step

abayer:
[JENKINS-26521](https://issues.jenkins-ci.org/browse/JENKINS-26521)

Replaces #18 

Many thanks to @ikedam for the original version of this. I've just resurrected it and tweaked a couple things.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
@@ -5,9 +5,14 @@ import com.google.common.util.concurrent.MoreExecutors;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Main;
 import hudson.Util;
+import hudson.console.ConsoleLogFilter;
+import hudson.console.LineTransformationOutputStream;
 import hudson.model.Result;
+import hudson.model.Run;
 import hudson.model.TaskListener;
 import java.io.IOException;
+import java.io.OutputStream;
+import java.io.Serializable;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledFuture;
@@ -31,30 +36,45 @@ public class TimeoutStepExecution extends AbstractStepExecutionImpl {
     private BodyExecution body;
     private transient ScheduledFuture<?> killer;
 
+    private long timeout = 0;
     private long end = 0;
+
+    /** Used to track whether this is timing out on inactivity without needing to reference {@link #step}. */
+    private boolean activity = false;
+
     /** whether we are forcing the body to end */
     private boolean forcible;
 
     TimeoutStepExecution(TimeoutStep step, StepContext context) {
         super(context);
         this.step = step;
+        this.activity = step.isActivity();
     }
 
     @Override
     public boolean start() throws Exception {
         StepContext context = getContext();
-        body = context.newBodyInvoker()
-                .withCallback(new Callback())
-                .start();
-        long now = System.currentTimeMillis();
-        end = now + step.getUnit().toMillis(step.getTime());
-        setupTimer(now);
+        BodyInvoker bodyInvoker = context.newBodyInvoker()
+                .withCallback(new Callback());
+
+        if (step.isActivity()) {
+            bodyInvoker = bodyInvoker.withContext(
+                    BodyInvoker.mergeConsoleLogFilters(
+                            context.get(ConsoleLogFilter.class),
+                            new ConsoleLogFilterImpl(new ResetCallback())
+                    )
+            );
+        }
+
+        body = bodyInvoker.start();
+        timeout = step.getUnit().toMillis(step.getTime());
+        resetTimer();
         return false;   // execution is asynchronous
     }
 
     @Override
     public void onResume() {
-        setupTimer(System.currentTimeMillis());
+        setupTimer(System.currentTimeMillis(), false);
     }
 
     private TaskListener listener() {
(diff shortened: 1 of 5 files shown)


jglick:
And what about #19?

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -31,30 +36,45 @@
     private BodyExecution body;
     private transient ScheduledFuture<?> killer;
 
+    private long timeout = 0;
     private long end = 0;
+
+    /** Used to track whether this is timing out on inactivity without needing to reference {@link #step}. */
```
jglick: Why not just delete the `step` field altogether at this point?

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -185,5 +229,47 @@ public String getShortDescription() {
...
+
+        private void logWritten() {
+            resetTimer();
+        }
+    }
+
+    private static class ConsoleLogFilterImpl extends ConsoleLogFilter implements Serializable {
+        private static final long serialVersionUID = 1L;
+
+        private final ResetCallback callback;
```
jglick: Easier to just make `ConsoleLogFilterImpl` non`static` and inline `logWritten`.

src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -139,6 +139,106 @@ public void evaluate() throws Throwable {
...
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(""
+                        + "node {\n"
+                        + "  timeout(time:5, unit:'SECONDS', activity: true) {\n"
+                        + "    echo 'NotHere';\n"
+                        + "    sleep 3;\n"
+                        + "    echo 'NotHereYet';\n"
+                        + "    sleep 3;\n"
```
jglick: Test could be flaky on loaded CI systems if you have only a 1s buffer.

svanoort:
The reliance on reading the logs here makes me rather nervous -- I can't put a finger on something clearly *wrong* with it per se, but it seems like it might open up some potential stability issues. 

We need to be really, really careful that we're not thrashing the system by forcing heavy log read activity.

abayer:
@svanoort - Yeah, I agree. This *works* but I don't love it. I'd made some vague efforts in the direction of implementing this before finding #18 - nothing else I could come up with would actually work. Particularly because the actual intent here really is to emulate build-timeout's "timeout if nothing new is written to the log" functionality - there really isn't much we can do there but sniff the logs in some way.

abayer:
@jglick - might it make sense to actually make "timeout on log inactivity" part of `DurableTask`, wait for [the `Controller.watch`](https://github.com/jenkinsci/durable-task-plugin/pull/60/files#diff-8c4fb36832416fbbbf56e1ac7747491eR301) api and piggyback on that with a `Handler` that pushes back "hey, log activity" notifications? Feels like there should be some way to sneak in before we're actually getting to the log-goes-to-master-or-wherever-else phase if we go straight into `DurableTask` and `DurableTaskStep`...

jglick:
@abayer perhaps. There could be some use cases for checking inactivity on miscellaneous long-running steps producing output—for example, if there is a step which tails a log from a grid engine. On the other hand I agree that `DurableTaskStep` is likely to account for the great majority of uses of this feature. https://github.com/jenkinsci/durable-task-plugin/pull/60 would not need to be involved, I think (nor https://github.com/jenkinsci/durable-task-plugin/pull/49, which also deals with something vaguely related); suffices for https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/63 (or indeed the current code) to be amended to send back a heartbeat signal.

But I do not think such a proposal really simplifies anything. Indeed what I meant by my comment was not that this PR is fundamentally incompatible with an external log sink—only that the _current implementation_ is. https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/21 means that the `ConsoleLogFilter` needs to be safely remotable.¹ That could be satisfied without any major rewrite (I think) by calling `Channel.export` to create a proxy for a heartbeat interface if and when the filter is remoted (as opposed to saved to `program.dat`). Thus, when new output appears, whether the `OutputStream` receiving it is running on the master (most steps) or on some agent (`sh` etc., with above changes), the heartbeat method gets called in the master and updates the running `TimeoutStepExecution`.

¹Unfortunately while we have a `SerializableOnlyOverRemoting` marker interface, we currently lack something like `SerializableAlsoOverRemoting`, so this is currently not mechanically checkable. CC @oleg-nenashev.

jglick:
> we're not thrashing the system by forcing heavy log read activity

I see no reason for worry. This PR does not “read” anything. It merely intercepts log writes. Yes the `LineTransformationOutputStream` adds a little bit of overhead, but I expect it to be very small, and after all this feature is opt-in.

abayer:
Where exactly would we be calling `Channel.export`?

abayer:
And is there an existing example of this sort of thing you could point me to?

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -134,7 +178,7 @@ private void cancel() {
             forcible = true;
             long now = System.currentTimeMillis();
             end = now + GRACE_PERIOD;
-            setupTimer(now);
+            resetTimer();
```
ikedam: Something strange.
`end` will be overridden soon in `resetTimer`.

henrik-koren:
@abayer , @ikedam  when do you think that this PR will be merge 
I am the initial requester of this change 
thanks

jglick:
> Where exactly would we be calling `Channel.export`?

From either `writeReplace` or `readResolve` of `ConsoleLogFilterImpl`, iff there is indeed an active `Channel` (thus serialization is over Remoting rather than to `program.dat`). `ResetCallback` would need to be made to implement an `interface` you could `export`. (This would of course supersede my suggestion to make `ConsoleLogFilterImpl` non`static`.)

abayer:
@jglick - So would we need to behave differently depending on whether we have a `Channel`, then? Is there any good example of something like this? `hudson.util.ProcessTree` is the first place I've stumbled over use of `Channel#export` - not sure if that's actually a *good* example, though.

jglick:
See https://github.com/abayer/workflow-basic-steps-plugin/pull/1.

abayer:
@jglick - Thanks! That was very similar to what I was hacking together anyway, just...well, simpler. =)

abayer:
Ping @svanoort

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -31,30 +37,45 @@
...
+        body = bodyInvoker.start();
+        timeout = step.getUnit().toMillis(step.getTime());
+        resetTimer();
         return false;   // execution is asynchronous
     }
 
     @Override
     public void onResume() {
-        setupTimer(System.currentTimeMillis());
+        setupTimer(System.currentTimeMillis(), false);
```
svanoort: @abayer I wonder a bit about what the implications are for not resetting activity-based timers at resume... if we're looking at logs, won't the timer trip if Jenkins has been down for a while, and it takes a little bit to get the new log data?

Then again we also trip normal timers at restart too I guess.

abayer: Yeah, I'm not sure what the ideal approach here would be, so I'd say let's go with this to start and change it in the future if needed. =)

bartoszkosiorek-tomtom: I have found issue with timeout activity, described in:
https://issues.jenkins.io/browse/JENKINS-58752

How additional logs could be taken from Jenkins, to investigate that issue?

PR #63, opened 2018-04-17
Stop using junit in metastep tests

abayer:
With newer junit plugin versions, these tests would have failed due to
`junit` now being an actual `Step`. This shows up right now in the
PCT. So let's get rid of the `junitResultArchiver` test, since it's
irrelevant, and change the `coreStepWithSymbol*` tests to use
`ArtifactArchiver` instead.

cc @reviewbybees


abayer:
Re-requested @svanoort review now that he's got approval rights. =)

PR #64, opened 2018-04-24
Add exception name in TimeoutStep help

alkuzad:
Using `Exception` is not best idea, better to catch this specific exception

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/steps/TimeoutStep/help.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/steps/TimeoutStep/help.html
@@ -1,6 +1,6 @@
 <div>
     Executes the code inside the block with a determined time out limit.
-    If the time limit is reached, an exception is thrown, which leads in aborting 
+    If the time limit is reached, an exception (FlowInterruptedException) is thrown, which leads in aborting 
     the build (unless it is caught and processed somehow).
     Unit is optional but defaults to minutes.
 </div>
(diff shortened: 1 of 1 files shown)


svanoort:
@dwnusbaum Yeah, I can understand being a little skeezed out by the implementation-specific version -- after doing some digging and some thought, I think it's okay here, and here's why:

* If you catch generic InterruptedException you can catch legitimate attempts to interrupt the thread or flow
* FlowInterruptedException is more specific than the generic exception someone might catch, which can result in catching things you might want to bubble up (more serious internal errors)

svanoort:
Thanks for the contribution, @alkuzad

jglick:
User scripts should not be referring to `FlowInterruptedException` explicitly. If there is a particular use case for `catch`ing discriminated exception types which might include Jenkins-specific types, this would be better handled by a new step, something akin to `catchError` I suppose.

jglick:
IOW, I guess a -0 on this. Fine to document what the behavior is in principle, but in this case the _presence_ of the documentation implies that you should be doing something which you should not.

alkuzad:
@jglick 

input was not designed as if/else but rather go/no-go tool
timeout was explictly designed as "after this time, build is screwed, abort" tool, not "maximum wait time"

yet, 2 try/catch later, they are changed to bow to management will :D

Clean way of doing this would be to create two new steps with desired behaviour and use them instead. However I lack Java skill and proper IDE to do such things, going around this problem with plain groovy was a lot easier.

PR #66, opened 2018-07-16
[JENKINS-52313] Add Base64 encoding to ReadFileStep and WriteFileStep for binary data

dwnusbaum:
See [JENKINS-52313](https://issues.jenkins-ci.org/browse/JENKINS-52313).

Adds support for a special encoding named "Base64" that can be used for binary files. Useful in tandem with jenkinsci/workflow-cps-global-lib-plugin#48 to copy binary resource files from shared libraries to the workspace.

~I'd prefer to use Java 8's Base64 API but the plugin is currently building against Java 7 so I am using commons-compress via Jenkins core.~ Baseline has been updated to Java 8 and Jenkins 2.60.3 to use the new Base64 API.

Also, I'm not sure if we want to add features to the stable branch or if that should be limited to the bleeding-edge branch, so just let me know if I need to rebase the PR.

@reviewbybees @jennbriden

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,8 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>2.7.3</jenkins.version>
-        <java.level>7</java.level>
+        <jenkins.version>2.60.3</jenkins.version>
+        <java.level>8</java.level>
         <workflow-step-api-plugin.version>2.13</workflow-step-api-plugin.version>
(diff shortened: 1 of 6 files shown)


svanoort:
@dwnusbaum I have zero objections if you bump up to Java 8 at this point and use the Base64 encoding API there - Jenkins 2.60 is over a year old, so we're fine to take up a requirement on it.  In fact, better to do that I think.

PR #69, opened 2018-09-07
JENKINS-53305 Workaround for email error

agentgonzo:
https://issues.jenkins-ci.org/browse/JENKINS-53305

_sometimes_ when emailing as part of a pipeline, we can get the error:
```
avax.mail.MessagingException: IOException while sending message;
  nested exception is:
	javax.activation.UnsupportedDataTypeException: no object DCH for MIME type multipart/mixed; 
	boundary="----=_Part_190_367171778.1535728877636"
	at com.sun.mail.smtp.SMTPTransport.sendMessage(SMTPTransport.java:1141)
	at javax.mail.Transport.send0(Transport.java:195)
	at javax.mail.Transport.send(Transport.java:124)
	at org.jenkinsci.plugins.workflow.steps.MailStep$MailStepExecution.run(MailStep.java:142)
	at org.jenkinsci.plugins.workflow.steps.MailStep$MailStepExecution.run(MailStep.java:128)
	at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution$1$1.call(SynchronousNonBlockingStepExecution.java:50)
	at hudson.security.ACL.impersonate(ACL.java:290)
	at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution$1.run(SynchronousNonBlockingStepExecution.java:47)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: javax.activation.UnsupportedDataTypeException: no object DCH for MIME type multipart/mixed; 
	boundary="----=_Part_190_367171778.1535728877636"
	at javax.activation.ObjectDataContentHandler.writeTo(DataHandler.java:896)
	at javax.activation.DataHandler.writeTo(DataHandler.java:317)
	at javax.mail.internet.MimeBodyPart.writeTo(MimeBodyPart.java:1476)
	at javax.mail.internet.MimeMessage.writeTo(MimeMessage.java:1772)
	at com.sun.mail.smtp.SMTPTransport.sendMessage(SMTPTransport.java:1099)
	... 12 more
```

This happens when `Thread.currentThread.getContextClassLoader()` is `null`. Why it should be `null` and what is setting it, I don't know. However, although this error only happens _sometimes_, I can recreate it every time by explicitly setting the class loader to `null` as a test.

This is a workaround, not a fix, so don't mind if it doesn't get merged, but am leaving it here for discussion or in case others are hitting the same problem. At the very least, if this doesn't get merged and you are hitting this  problem, you can do a custom build of the plugin using this PR and hopefully get less failures from your Jenkins instance.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
@@ -30,6 +30,10 @@ import hudson.model.TaskListener;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import jenkins.model.Jenkins;
 import jenkins.plugins.mailer.tasks.MimeMessageBuilder;
 
 import org.apache.commons.lang.StringUtils;
@@ -48,6 +52,8 @@ import javax.mail.internet.MimeMessage;
  * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
  */
 public class MailStep extends Step {
+
+    private static final Logger LOGGER = Logger.getLogger(MailStep.class.getName());
 
     private String charset;
 
@@ -139,6 +145,14 @@ public class MailStep extends Step {
         @Override
         protected Void run() throws Exception {
             MimeMessage mimeMessage = buildMimeMessage();
+            // JENKINS-53305 - contextClassLoader can _sometimes_ be null, which causes the obtuse error message
+            // javax.activation.UnsupportedDataTypeException: no object DCH for MIME type multipart/mixed;
+            // I can't see why this would ever be valid and don't know the cause, but if we encounter it, then set
+            // it to the uberclass loader
+            if (Thread.currentThread().getContextClassLoader() == null) {
+                LOGGER.log(Level.WARNING, "contextClassLoader is null - re-setting to uberClassLoader");
+                Thread.currentThread().setContextClassLoader(Jenkins.getInstance().getPluginManager().uberClassLoader);
+            }
             Transport.send(mimeMessage);
             return null;
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
@@ -148,10 +148,10 @@ public class MailStep extends Step {
             // JENKINS-53305 - contextClassLoader can _sometimes_ be null, which causes the obtuse error message
             // javax.activation.UnsupportedDataTypeException: no object DCH for MIME type multipart/mixed;
             // I can't see why this would ever be valid and don't know the cause, but if we encounter it, then set
-            // it to the uberclass loader
+            // it to the Jenkins loader
             if (Thread.currentThread().getContextClassLoader() == null) {
-                LOGGER.log(Level.WARNING, "contextClassLoader is null - re-setting to uberClassLoader");
-                Thread.currentThread().setContextClassLoader(Jenkins.getInstance().getPluginManager().uberClassLoader);
+                LOGGER.log(Level.WARNING, "contextClassLoader is null - re-setting to Jenkins.class.classLoader");
+                Thread.currentThread().setContextClassLoader(Jenkins.class.getClassLoader());
             }
             Transport.send(mimeMessage);
             return null;


dwnusbaum:
@agentgonzo Just a quick FYI, the `stable` branch should be used as the base of most PRs, `master` is just for changes that need a new Jenkins core baseline (we have two release lines in the plugin right now). We should probably make that clearer with a PR template or something.

agentgonzo:
@dwnusbaum - Ah, OK. I thought I'd messed up when setting it up - I have changed it back now.

svanoort:
Sorry for the notification noise, that was supposed to be an approval comment.

agentgonzo:
@svanoort - I came across those lines during my investigation. I could see that the `workflow-cps-plugin` _was_ setting the contextClassLoader to `null` _on some occasions_, but I could trace that back to the fact that preceeding those calls, but that was due to the fact that `ORIGINAL_CONTEXT_CLASSLOADER` was already `null` here: https://github.com/jenkinsci/workflow-cps-plugin/blob/0e4c25f8d7b84470aa523491e29933db3b3df588/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java#L83-L84 - which means that for its part, the CPS plugin is behaving 'as intended' and is is something else that is erroneously (in my eyes) setting the contextClassLoader on a thread (which gets inherited by the new `ThreadContext`) to `null`

agentgonzo:
I feel the same way. I would prefer to understand the root cause, but I
didn't the best part of a week trying to track it down and couldn't.
There's nothing we can from within this plugin to get more information on
the root cause because the thread that this code runs in is
spawned/inherited from the CPS workflow plugin and even in the, the
contectClassLoader seems to be null when we first ever the plugins
functional code.

On the other hand, this is a workaround with no side effects that _does_
make a problem 'go away', even if it's not ideal.

If we could find a way of reliably recreating the issue in a test/local
environment, then we could make progress, but I couldn't do that.

It's up to you whether to merge or not. I've done a local build and
deployed that version of the plugin to 'fix' our infra - others can do the
same if this doesn't get merged.

No solution at this point is ideal

src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
```diff
@@ -139,6 +145,14 @@ public StepExecution start(StepContext context) throws Exception {
         @Override
         protected Void run() throws Exception {
             MimeMessage mimeMessage = buildMimeMessage();
+            // JENKINS-53305 - contextClassLoader can _sometimes_ be null, which causes the obtuse error message
+            // javax.activation.UnsupportedDataTypeException: no object DCH for MIME type multipart/mixed;
+            // I can't see why this would ever be valid and don't know the cause, but if we encounter it, then set
+            // it to the uberclass loader
+            if (Thread.currentThread().getContextClassLoader() == null) {
+                LOGGER.log(Level.WARNING, "contextClassLoader is null - re-setting to uberClassLoader");
+                Thread.currentThread().setContextClassLoader(Jenkins.getInstance().getPluginManager().uberClassLoader);
```
jglick: No, this is _not_ the normal `contextClassLoader` in Jenkins (surprisingly). Rather use `Jenkins.class.getClassLoader()`.

agentgonzo:
@jglick - I've changed the class loader to `Jenkins.class.getClassLoader()` now. As mentioned, I don't have any more time to investigate this issue wrt `ClassLoaderSanityThreadFactory` but if I get the time in the future I will take a look at it. For the moment, I'll leave this PR in case it's a valid workaround for others hitting this issue.

agentgonzo:
>you need to request me as a reviewer again since I cannot keep up with notifications generally

@jglick - I don't have permissions on this repo to request reviewers :-(

agentgonzo:
@rsandell - could you request @jglick as a reviewer please?

rsandell:
@agentgonzo already did two days ago.
@jglick ping? Do you still want to take another look?

rsandell:
No, I removed review and then re-requested it :) He is still in the reviewers list.

PR #70, opened 2018-09-10
Fix file leak in WriteFileStep when using Base64 encoding

dwnusbaum:
Fixes a file leak specific to Base64 encoding introduced by #66. Similar to jenkinsci/workflow-cps-global-lib-plugin#52.

@reviewbybees


svanoort:
@dwnusbaum Please could you merge and go ahead to cut a release?  Thanks!

PR #74, opened 2018-10-08
[JENKINS-54078] JEP-210 compatibility

jglick:
Demonstrates that https://github.com/abayer/workflow-basic-steps-plugin/pull/1 was a necessary amendment to https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/62 in order to make `timeout` work properly with remote durable tasks as of JEP-210. If you disable that fix using

```diff
diff --git a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
index 20f2f21..0b4933c 100644
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
@@ -250,11 +250,6 @@ public class TimeoutStepExecution extends AbstractStepExecutionImpl {
             this.callback = callback;
         }
 
-        private Object writeReplace() {
-            Channel ch = Channel.current();
-            return ch == null ? this : new ConsoleLogFilterImpl(ch.export(ResetCallback.class, callback));
-        }
-
         @Override
         public OutputStream decorateLogger(@SuppressWarnings("rawtypes") Run build, final OutputStream logger)
                 throws IOException, InterruptedException {
```

then the build fails as expected:

```
[Pipeline] sh
[p] Running shell script
[Pipeline] }
[Pipeline] // timeout
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
java.io.NotSerializableException: hudson.model.Executor
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
	at …
	at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:348)
	at hudson.remoting.UserRequest._serialize(UserRequest.java:264)
	at hudson.remoting.UserRequest.serialize(UserRequest.java:273)
	at hudson.remoting.UserRequest.<init>(UserRequest.java:101)
	at hudson.remoting.Channel.call(Channel.java:953)
	at hudson.Launcher$RemoteLauncher.launch(Launcher.java:1052)
	at hudson.Launcher$ProcStarter.start(Launcher.java:449)
	at org.jenkinsci.plugins.durabletask.BourneShellScript.launchWithCookie(BourneShellScript.java:185)
	at org.jenkinsci.plugins.durabletask.FileMonitoringTask.launch(FileMonitoringTask.java:98)
	at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.start(DurableTaskStep.java:266)
	at org.jenkinsci.plugins.workflow.cps.DSL.invokeStep(DSL.java:270)
	at …
Caused: java.io.IOException: Unable to serialize hudson.Launcher$RemoteLaunchCallable@2b4d80f
	at hudson.remoting.UserRequest.serialize(UserRequest.java:275)
	at hudson.remoting.UserRequest.<init>(UserRequest.java:101)
	at hudson.remoting.Channel.call(Channel.java:953)
	at hudson.Launcher$RemoteLauncher.launch(Launcher.java:1052)
	at hudson.Launcher$ProcStarter.start(Launcher.java:449)
	at org.jenkinsci.plugins.durabletask.BourneShellScript.launchWithCookie(BourneShellScript.java:185)
	at org.jenkinsci.plugins.durabletask.FileMonitoringTask.launch(FileMonitoringTask.java:98)
	at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.start(DurableTaskStep.java:266)
	at org.jenkinsci.plugins.workflow.cps.DSL.invokeStep(DSL.java:270)
	at …
Finished: FAILURE
```

Some more discussion: https://github.com/jenkinsci/workflow-support-plugin/pull/15#discussion_r210586067

----

And then amends that change for [JENKINS-54078](https://issues.jenkins-ci.org/browse/JENKINS-54078) because it turns out (as I just happened to learn yesterday in https://github.com/jenkinsci/pipeline-log-fluentd-cloudwatch-plugin/pull/8/commits/f66f63810e16929f5dd3f61ca2489df492e9ef8e!) that `Channel.export` is pretty much useless; for some reason I have yet to track down, `FilePath` calls from the Git plugin (but not those from, say, the `sh` step) trigger collection of references.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,8 @@
         <workflow-support-plugin.version>2.21-beta-1</workflow-support-plugin.version>
         <workflow-api-plugin.version>2.30-beta-1</workflow-api-plugin.version>
         <useBeta>true</useBeta>
+        <git-plugin.version>4.0.0-beta3</git-plugin.version>
+        <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
     </properties>
     <dependencies>
         <dependency>
@@ -204,13 +206,33 @@
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>git</artifactId>
+            <version>${git-plugin.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>scm-api</artifactId>
+            <version>${scm-api-plugin.version}</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>git</artifactId>
+            <version>${git-plugin.version}</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <dependencyManagement>
         <dependencies>
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>
                 <artifactId>scm-api</artifactId>
-                <version>2.2.6</version>
+                <version>${scm-api-plugin.version}</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -129,7 +129,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-durable-task-step</artifactId>
-            <version>2.22-rc713.c51e25cf0eef</version> <!-- TODO https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/75 -->
+            <version>2.24</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.19</version>
+        <version>3.25</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -67,9 +67,9 @@
         <jenkins.version>2.121.1</jenkins.version>
         <java.level>8</java.level>
         <workflow-step-api-plugin.version>2.15</workflow-step-api-plugin.version>
-        <workflow-cps-plugin.version>2.32</workflow-cps-plugin.version>
-        <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
-        <workflow-api-plugin.version>2.28</workflow-api-plugin.version>
+        <workflow-cps-plugin.version>2.58-beta-1</workflow-cps-plugin.version>
+        <workflow-support-plugin.version>2.21-beta-1</workflow-support-plugin.version>
+        <workflow-api-plugin.version>2.30-beta-1</workflow-api-plugin.version>
         <useBeta>true</useBeta>
     </properties>
     <dependencies>
@@ -98,7 +98,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.14</version>
+            <version>1.17</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -121,13 +121,13 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.10</version>
+            <version>2.26-beta-1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-durable-task-step</artifactId>
-            <version>2.3</version>
+            <version>2.22-rc713.c51e25cf0eef</version> <!-- TODO https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/75 -->
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -182,7 +182,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.28</version>
+            <version>1.46</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -205,4 +205,13 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.jenkins-ci.plugins</groupId>
+                <artifactId>scm-api</artifactId>
+                <version>2.2.6</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -67,9 +67,9 @@
         <jenkins.version>2.121.1</jenkins.version>
         <java.level>8</java.level>
         <workflow-step-api-plugin.version>2.15</workflow-step-api-plugin.version>
-        <workflow-cps-plugin.version>2.58-beta-1</workflow-cps-plugin.version>
-        <workflow-support-plugin.version>2.21-beta-1</workflow-support-plugin.version>
-        <workflow-api-plugin.version>2.30-beta-1</workflow-api-plugin.version>
+        <workflow-cps-plugin.version>2.58</workflow-cps-plugin.version>
+        <workflow-support-plugin.version>2.21</workflow-support-plugin.version>
+        <workflow-api-plugin.version>2.30</workflow-api-plugin.version>
         <useBeta>true</useBeta>
         <git-plugin.version>4.0.0-beta3</git-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
@@ -123,7 +123,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.26-beta-1</version>
+            <version>2.26</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
@@ -10,17 +10,22 @@ import hudson.console.LineTransformationOutputStream;
 import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+import hudson.remoting.Callable;
 import hudson.remoting.Channel;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.List;
+import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import jenkins.model.CauseOfInterruption;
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.19</version>
+        <version>3.25</version>
```
jglick: Various changes to pick up JEP-210 binaries.

pom.xml
```diff
@@ -121,13 +121,13 @@
...
             <artifactId>workflow-job</artifactId>
-            <version>2.10</version>
+            <version>2.26-beta-1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-durable-task-step</artifactId>
-            <version>2.3</version>
+            <version>2.22-rc713.c51e25cf0eef</version> <!-- TODO https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/75 -->
```
jglick: https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/75, to get `ABORTED` rather than `FAILURE` in the test.

Note that this build is actually _older_ than the 2.22 on the update center, but it does contain https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/63 and the subsequent commits are uninteresting.

src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -239,6 +241,24 @@ public void evaluate() throws Throwable {
...
+    @Test
+    public void activityRemote() {
+        assumeFalse(Functions.isWindows()); // TODO create analogue using bat
+        story.then(r -> {
+            r.createSlave();
+            WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("" +
+                     "node('!master') {\n" +
+                     "  timeout(time:5, unit:'SECONDS', activity: true) {\n" +
+                     "    sh 'set +x; echo NotHere; sleep 3; echo NotHereYet; sleep 3; echo JustHere; sleep 10; echo ShouldNot'\n" +
```
jglick: Cannot use `echo` steps as in the other tests, because we need to demonstrate serialization of the `ConsoleLogFilter` impl, which only happens if a nested step actually remotes its `TaskListener`.

dwnusbaum:
Looks like the test failure is because `EnvStepTest` needs to be updated to handle the new log format now that the parallel branch name isn't shown in the logs: 

https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/8d0a946b0234f843c203271d02f0df4abdc35ebe/src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java#L103

pom.xml
```diff
@@ -67,9 +67,9 @@
         <jenkins.version>2.121.1</jenkins.version>
         <java.level>8</java.level>
         <workflow-step-api-plugin.version>2.15</workflow-step-api-plugin.version>
-        <workflow-cps-plugin.version>2.32</workflow-cps-plugin.version>
-        <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
-        <workflow-api-plugin.version>2.28</workflow-api-plugin.version>
+        <workflow-cps-plugin.version>2.58-beta-1</workflow-cps-plugin.version>
```
dwnusbaum: These can be updated to non-beta versions now.

dwnusbaum (approved):

pom.xml
jglick: Remind me to do so once https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/75 is in a release.

dwnusbaum: Ok, if that PR isn't released by next Friday I will release it and ping you then.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -230,29 +239,41 @@ public String getShortDescription() {
...
+    private static final class ResetTimer extends SlaveToMasterCallable<Void, RuntimeException> {
 
-    private class ResetCallbackImpl implements ResetCallback {
         private static final long serialVersionUID = 1L;
-        @Override public void logWritten() {
-            resetTimer();
+
+        private final String id;
+
+        ResetTimer(String id) {
```
dwnusbaum: It doesn't matter that the slave provides this value from a security standpoint, right? Even if bogus ids were passed, and the agent magically guessed a UUID for another build correctly, it would just reset the activity timer, which doesn't really matter.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -230,29 +239,41 @@ public String getShortDescription() {
...
 
-        ConsoleLogFilterImpl(ResetCallback callback) {
-            this.callback = callback;
+        ConsoleLogFilterImpl(String id) {
+            this.id = id;
         }
 
-        private Object writeReplace() {
-            Channel ch = Channel.current();
-            return ch == null ? this : new ConsoleLogFilterImpl(ch.export(ResetCallback.class, callback));
```
dwnusbaum: Maybe the issue is that only the marshaller holds the reference to the exported object, so after the object is completely serialized it is free to be garbage collected?

dwnusbaum (dismissed):
Looks fine to me, although it would be nice to understand the root cause.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
jglick: AFAICT the issue was that [this code](https://github.com/jenkinsci/remoting/blob/81d1beda0dd2cbc6ca845d1bc812e45c553ae980/src/main/java/hudson/remoting/UserRequest.java#L299) released all exported objects when a `Callable` finished, insofar as those objects had originally been exported within the dynamic scope of that `Callable`’s serialization. Or something like this. I could not follow exactly what the logic was, and the documentation was scant. At any rate, the revised strategy with the `SlaveToMasterCallable` is simpler to reason about explicitly.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
jglick: Right, it could not do much damage.

jglick:
I intend to amend 0918918 to be more secure. Also the CI seems to be broken and I would like to get a clean build.

jglick:
Having difficulty coming up with a convincing test case for the upgrade scenario. Was playing with something like

```java
@Issue("JENKINS-54078")
@LocalData
@Test public void activityUpgrade() {
    story.then(r -> {
        r.createSlave();
        /* Setup, with workflow-job reverted to 2.25 and src/main/ from workflow-basic-steps 2.11:
        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
        p.setDefinition(new CpsFlowDefinition(
            "node('!master') {\n" +
            "  timeout(time: 5, unit: 'SECONDS', activity: true) {\n" +
            "    remotePrint 'NotHere'\n" +
            "    sleep 3\n" +
            "    remotePrint 'NotHereYet'\n" +
            "    sleep 3\n" +
            "    remotePrint 'JustHere'\n" +
            "    while (true) {\n" +
            "      sleep 10\n" +
            "      remotePrint 'ShouldNot'\n" +
            "    }\n" +
            "  }\n" +
            "}\n", true));
        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
        r.waitForMessage("“JustHere” from agent", b);
        r.waitForMessage("Sleeping for 10 sec", b);
        Runtime.getRuntime().halt(1);
        */
        WorkflowRun b = r.jenkins.getItemByFullName("p", WorkflowJob.class).getBuildByNumber(1);
        r.assertBuildStatus(Result.ABORTED, r.waitForCompletion(b));
        r.assertLogContains("Cancelling nested steps due to timeout", b);
        /* Actually we get this, once, because the sleep expired:
        r.assertLogNotContains("ShouldNot", b);
        */
        b.getLogText().writeRawLogTo(0, System.out);
    });
}
// TODO adapted from workflow-job; consider pushing into tests JAR of workflow-support for reuse
public static final class RemotePrintStep extends Step {
    public final String message;
    @DataBoundConstructor public RemotePrintStep(String message) {
        this.message = message;
    }
    @Override public StepExecution start(StepContext context) throws Exception {
        return new Execution(context, message);
    }
    private static final class Execution extends SynchronousNonBlockingStepExecution<Void> {
        private final String message;
        Execution(StepContext context, String message) {
            super(context);
            this.message = message;
        }
        @Override protected Void run() throws Exception {
            return getContext().get(Node.class).getChannel().call(new PrintCallable(getContext().get(TaskListener.class), message));
        }
    }
    private static final class PrintCallable extends MasterToSlaveCallable<Void, RuntimeException> {
        private final TaskListener listener;
        private final String message;
        PrintCallable(TaskListener listener, String message) {
            this.listener = listener;
            this.message = message;
        }
        @Override public Void call() throws RuntimeException {
            listener.getLogger().println("“" + message + "” from " + (JenkinsJVM.isJenkinsJVM() ? "master" : "agent"));
            return null;
        }
    }
    @TestExtension public static final class DescriptorImpl extends StepDescriptor {
        @Override public Set<? extends Class<?>> getRequiredContext() {
            return ImmutableSet.of(Node.class, TaskListener.class);
        }
        @Override public String getFunctionName() {
            return "remotePrint";
        }
    }
}
```

However even _before_ the upgrade it does not work unless I also

```diff
diff --git a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
index 20f2f21..76c0154 100644
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
@@ -75,7 +75,11 @@ public class TimeoutStepExecution extends AbstractStepExecutionImpl {
 
     @Override
     public void onResume() {
-        setupTimer(System.currentTimeMillis(), false);
+        if (activity) {
+            resetTimer();
+        } else {
+            setupTimer(System.currentTimeMillis(), false);
+        }
     }
 
     private TaskListener listener() {
```

because `end` is set to be however many seconds since the last activity _before_ the restart, and so as soon as the build resumes the `timeout` block is cancelled. I am not sure how to reproduce this problem, if it is a problem, in `activityRestart`. Probably it is necessary to have a single (atom) step running across the upgrade and periodically emitting output, so that we do not see the effect of the Pipeline step notes resetting the timer; `sh` cannot be used for this because it is not easily resumed in `@LocalData`.

Related to that, there seems to be no existing coverage for the case that activity _is_ seen after a restart and the build succeeds.

Anyway, I think for now I am giving up on a test case for the upgrade, and will just write the probable fix based on reasoning.

dwnusbaum (approved):
The fix looks good to me, and I did some manual testing across upgrades which worked as expected.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
jglick: The compatibility issue was sorted out, I think.

dwnusbaum: Yes, I manually confirmed that there are no problems with upgrading the plugin while builds are running inside of a `timeout(activity:true...)` step, so I'll mark this thread as resolved.

src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -239,6 +244,42 @@ public void evaluate() throws Throwable {
         });
     }
 
+    @Test
+    public void activityRemote() {
```
jglick, line 248: I think this does not actually test anything as of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/84.

PR #75, opened 2018-10-29
[JENKINS-54323] Add setBuildInfo step

dwnusbaum:
See [JENKINS-54323](https://issues.jenkins-ci.org/browse/JENKINS-54323). Complements https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/71. 

This PR adds a dedicated step named `setBuildInfo` (I don't feel too strongly about that name, so let me know if you'd prefer something else) for setting `Run.description`, `Run.displayName`, `Run.keepLog`, and `Run.result`. This step would get rid of one of the use cases that currently requires `RunWrapper`.

Right now, I've created a single step that takes a specified `Describable` in its constructor:

    setBuildInfo(description('description here'))
    setBuildInfo(displayName('display name here'))
    setBuildInfo(keepLog(true))
    setBuildInfo(result('FAILURE'))

The other approach that would make sense to me would be to have a dedicated step for each field:

    setBuildDescription('description here')
    setBuildDisplayName('display name here')
    setBuildKeepLog(true)
    setBuildResult('FAILURE')

I chose the describable approach because I initially allowed a list to be passed to the step, but that didn't seem very likely to be useful in practice, so I simplified it, at which point a dedicated step for each field seems like the better approach. I'm putting this up as-is to see what people think, so let me know if you have a strong preference for either implementation.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/SetBuildInfoStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/SetBuildInfoStep.java
@@ -1,0 +1,207 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2018 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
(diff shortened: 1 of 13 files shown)


dwnusbaum:
@abayer Yeah, I wasn't sure if we wanted that to be handled along with the API step (i.e. as shims on top of specific API queries), or if we should just implement the common ones directly in the obvious way. Thoughts?

abayer:
Oh yeah, forgot about that. Well...hmm. I'd say that if we're exposing `set` for these things, `get` as well via the same approach would only make sense.

jglick:
The `api` step already handles the get case (for both the current build and others), so I see no reason to bother with that. For compatibility, the `currentBuild` shim in off-master would manage gets.

src/main/java/org/jenkinsci/plugins/workflow/steps/SetBuildInfoStep.java
```diff
@@ -0,0 +1,207 @@
...
+
+        Execution(SetBuildInfoStep step, StepContext context) {
+            super(context);
+            this.step = step;
+        }
+
+        @Override
+        public Void run() throws Exception {
+            Run r = getContext().get(Run.class);
+            if (r != null) {
```
jglick: It is required context; you need not and should not check for null.

src/main/java/org/jenkinsci/plugins/workflow/steps/SetBuildInfoStep.java
```diff
@@ -0,0 +1,207 @@
...
+            return "Modify properties of the current build";
+        }
+
+        @Override
+        public Set<? extends Class<?>> getRequiredContext() {
+            return Collections.singleton(Run.class);
+        }
+    }
+
+    @Restricted(NoExternalUse.class)
```
jglick: Or could be made an `ExtensionPoint` if the need arises.

src/main/java/org/jenkinsci/plugins/workflow/steps/SetBuildInfoStep.java
```diff
@@ -0,0 +1,207 @@
...
+    public static abstract class RunPropertySetter extends AbstractDescribableImpl<RunPropertySetter> {
+        public abstract void set(Run r) throws Exception;
```
jglick: `rawtypes`

jglick:
> a dedicated step for each field seems like the better approach

No particular preference. I suppose the dedicated step is less verbose for users.

src/main/java/org/jenkinsci/plugins/workflow/steps/SetBuildInfoStep.java
abayer: IDEs tend to glare at you if you don't. =)

dwnusbaum:
No plans to pick this back up in the near future.

PR #80, opened 2019-01-23
[JENKINS-55612] Add optional label to echo step

bitwiseman:
[JENKINS-55612](https://issues.jenkins-ci.org/browse/JENKINS-55612)

This is a minimal implementation for this issue, including some basic tests.     

Based on this PR: https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/93

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/EchoStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/EchoStep.java
@@ -27,7 +27,11 @@ package org.jenkinsci.plugins.workflow.steps;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.model.TaskListener;
+import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
+import org.jenkinsci.plugins.workflow.actions.LabelAction;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import java.util.Collections;
 import java.util.Set;
 
@@ -36,10 +40,18 @@ import java.util.Set;
  */
 public class EchoStep extends Step {
 
-    private final String message;
+    private static final int MAX_LABEL_LENGTH = 100;
+
+    private String message;
+    private String label;
 
     @DataBoundConstructor
     public EchoStep(String message) {
+        this.message = message;
+    }
+
+    @DataBoundSetter
+    public void setMessage(String message) {
         this.message = message;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/EchoStep.java
```diff
@@ -36,19 +40,43 @@
...
+    }
+
     public String getMessage() {
         return message;
     }
 
+    @DataBoundSetter
+    public void setLabel(String label) {
+        if (label == null)
+            throw new IllegalArgumentException();
```
dwnusbaum: It feels a little awkward to me to throw an exception here. On the one hand it makes sense to fail fast, but I don't think anyone should be calling this method directly except for Stapler and maybe devs in a unit test, so perhaps adding an `@Nonnull` annotation would be good enough, and you still have to have the null check in `start` anyway for builds deserialized from before the label existed so it seems like a bit of a wash. Maybe you could add a `doCheckLabel` method to the descriptor to do some validation for anyone using the pipeline syntax page, but I'm not sure whether that works well with optional `@DataBoundSetter`s.

bitwiseman: @dwnusbaum
Yeah, since I check for null in the `start` this is overkill.  I like the `@NonNull` annotation for self-documenting code.

bitwiseman:
@dwnusbaum 
As to your general comment, are you pointing to a more general issue that I should file and come back to?  Or are you saying I need to change something in this PR? 
...

Ah, wait. 
I'm looking at the issue description ( https://issues.jenkins-ci.org/browse/JENKINS-55612) again. What the user wants differs from what they end up asking for. 

1. They want echo to still display a meaningful label even when the message is long. 
2. If there isn't some good way to do this, they'd like to add a `label` field they can set manually.  

Back to the drawing board for a minute.

bitwiseman:
@dwnusbaum  I've updated this PR with a version that truncate the argument string to the first line of the message but also adds the label field. I haven't run this up in a local blue ocean yet to verify the behavior.

jglick:
This PR exemplifies the reason I thought JENKINS-55410 was a bad idea to begin with—it just leads to endless creeping requests to add `label` to more and more steps, ad-hoc. I still feel that it would be a better design to use `stage` for the purpose for which it was designed¹: associating a human-readable label with a block of script. If some UI like Blue Ocean does not currently honor nested `stage`s well, fix that UI rather than forcing scripts to work around it. IMO.

And what really is the use case here?

> if the message for echo step is too long, it solely prints **Print Message** in pipeline view

This behavior is not fixed in stone. It is just the current design of `ArgumentsAction[Impl]` that long arguments are replaced by `OVERSIZE_VALUE` and the original text completely lost, rather than simply being truncated after newlines, too many characters, etc. so that you would see something like

> **PhpStorm inspection results: …**

without needing to change your script. I actually argued for such a behavior during the design, but it was deferred: https://github.com/jenkinsci/workflow-api-plugin/pull/26#discussion_r114809477

----
¹ Actually my original choice of step name was `label`, leaving the original behavior of `stage` untouched and deprecating it, but @kohsuke overrode me on that.

PR #81, opened 2019-02-17
Make it possible to detect source of timeout

rudolfwalter:
Make it possible to detect whether a FlowInterruptedException was caught inside or outside the timeout block that generated it. Workaround for [JENKINS-51928](https://issues.jenkins-ci.org/browse/JENKINS-51928) – user code can now decide to rethrow or not.

I made the constructor of `ExceededTimeout` private. In theory this is a breaking change; but I presumed it's improbable that external code was using it. Let me know if you agree.

Also, I wanted to add another check to the `serialForm` test to make sure future code stays compatible with this form as well, but I could not figure out how to do it. I tried creating a job similar to what the test already uses, and copy out the files (`program.dat` etc) while it is running, but for some reason it appears not to resume correctly under testing. Specifically, it says `No need to sleep any longer` under a real Jenkins as expected, but says `Sleeping for 2 min 0 sec` and gets stuck when resumed in the test case. Guidance would be appreciated.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
@@ -35,6 +35,7 @@ import java.util.concurrent.TimeUnit;
 import jenkins.model.CauseOfInterruption;
 import jenkins.model.InterruptedBuildAction;
 import jenkins.plugins.git.GitSampleRepoRule;
+import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepAtomNode;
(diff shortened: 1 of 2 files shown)


ppRadu:
This would be useful for us as well, @dwnusbaum could you take a look?

src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -421,4 +422,95 @@ private Execution(StepContext context) {
...
+                ScriptApproval.get().approveSignature("method org.jenkinsci.plugins.workflow.steps.TimeoutStepExecution$ExceededTimeout isInsideTimeoutBlock");
+                WorkflowRun b;
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+
+                //inside
+                p.setDefinition(new CpsFlowDefinition(""
+                    + "timeout(time: 2, unit: 'SECONDS') {\n"
+                    + "  try { sleep 10 }\n"
+                    + "  catch(e) {\n"
+                    + "    echo e.causes[0].isInsideTimeoutBlock() ? 'GOOD' : 'BAD'\n"
```
jglick, line 439: idiomatically
```suggestion
                    + "    echo e.causes[0].insideTimeoutBlock ? 'GOOD' : 'BAD'\n"
```

PR #83, opened 2019-04-18
[JENKINS-45579] Add steps to set unstable status and enhance catchError

dwnusbaum:
See https://github.com/jenkinsci/workflow-api-plugin/pull/91 and [JENKINS-45579](https://issues.jenkins-ci.org/browse/JENKINS-45579). (I am intentionally not updating the status of that ticket or assigning it to myself until I am ready to merge/release all relevant changes to avoid giving users premature hope of a fix and so that I have clear answers for what will and won't work and when any fixes will be released.)

This PR adds a step named `unstable` to mark a specific `FlowNode` as unstable. This can be used by higher-level APIs that visualize entire stages/branches to mark a specific stage/branch as unstable. The PR also adds a step named `warnError` that is like `catchError` but marks the build and `warnError` step as unstable if an exception is caught. Finally, the PR adds some options to `catchError` that allow it to be used to set arbitrary build and step results, to have a custom message, and to handle or rethrow `FlowInterruptedException`.


dwnusbaum:
Leaving this in draft status until I know exactly what API we want to propose.

jglick:
> a step named something like `warnError` that is like `catchError` but marks the build as unstable, since that could be used in Declarative

Interesting thought.

src/main/java/org/jenkinsci/plugins/workflow/steps/UnstableStep.java
```diff
@@ -0,0 +1,91 @@
...
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class UnstableStep extends Step {
+
+    private final String message;
+
+    @DataBoundConstructor
+    public UnstableStep(String message) {
+        message = Util.fixEmptyAndTrim(message);
+        if (message == null) {
```
jglick: I think you should need to pass a message.

src/main/java/org/jenkinsci/plugins/workflow/steps/UnstableStep.java
```diff
@@ -0,0 +1,91 @@
...
+    @Extension
+    public static class DescriptorImpl extends StepDescriptor {
+        @Override
+        public String getFunctionName() {
+            return "unstable";
+        }
+
+        @Override
+        public Set<? extends Class<?>> getRequiredContext() {
+            // Run, FlowNode, and TaskListener are always available.
```
jglick: Please include them nonetheless.

src/main/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStep.java
```diff
@@ -53,59 +125,98 @@ public final class CatchErrorStep extends Step {
...
+                    Result buildResult = options.getBuildResultOnError();
+                    Result stepResult = options.getStepResultOnError();
                     if (t instanceof AbortException) {
                         listener.error(t.getMessage());
                     } else if (t instanceof FlowInterruptedException) {
                         FlowInterruptedException fie = (FlowInterruptedException) t;
                         fie.handle(context.get(Run.class), listener);
-                        r = fie.getResult();
+                        buildResult = fie.getResult();
+                        stepResult = fie.getResult();
```
jglick, line 210: I am confused by this logic. If I kept `catchInterruptions` on, and set `buildResult` and `stageResult` to null, I would expect that anything that happens inside the `catchError` step could not fail the build. Yet, apparently, a `FlowInterruptedException` can still fail the build. Am I missing something?

dwnusbaum: I do not remember if this was intentional or not. My only guess it that the idea was to preserve the old behavior where `FlowInterruptedException.result` was always used to set the build result. If it does not break existing tests then I am not opposed to deleting these lines, or perhaps it could make sense to continue to use `FlowInterruptedException.result` but only when it `Result.isBetterThan` `getBuildResultOnError`/`getStepResultOnError`.

jglick: Not sure. In the meantime, the workaround for the record is to just use `try`…`catch`.

PR #85, opened 2019-05-29
[JENKINS-57537] Change type of buildResult and stageResult to String

dwnusbaum:
See [JENKINS-57537](https://issues.jenkins-ci.org/browse/JENKINS-57537). Follow-up to #83. 

I think having the type of the fields here be `Result` was a bad idea in the first place (As far as I can tell the only reason it even worked is because [Result registers a custom Stapler converter that will convert anything into a Result using `Object.toString`](https://github.com/jenkinsci/jenkins/blob/587da142a8ee6214f1c254323f03b61f4749051f/core/src/main/java/hudson/model/Result.java#L213-L217)?!), so although we might be able to fix the reported issue with tweaks to Declarative's validation, I think it makes more sense to just change the type of these fields to `String`.

I created a test to verify that this fixes the issue in Declarative [here](https://github.com/jenkinsci/pipeline-model-definition-plugin/commit/787d3504c27f36a67478c052011d5f534d474553), but I don't think it makes sense to actually add that test to Declarative since this PR fixes the issue purely in workflow-basic-steps and it's not something that seems likely to regress.


src/main/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStep.java
```diff
@@ -114,6 +116,25 @@ public void setCatchInterruptions(boolean catchInterruptions) {
         this.catchInterruptions = catchInterruptions;
     }
 
+    private void readObject(ObjectInputStream ois) throws ClassNotFoundException, IOException {
+        ObjectInputStream.GetField fields = ois.readFields();
+        message = (String) fields.get("message", null);
+        catchInterruptions = fields.get("catchInterruptions", true);
+        // Previously, the types of buildResult and stageResult were Result rather than String, so we handle either type.
```
dwnusbaum: This is a bit of a mess. We could mark the plugin as incompatible and break any Pipeline that runs across the upgrade using `catchError`, but it seems better to try to maintain compatibility.

PR #89, opened 2019-06-13
Clarify that stashes are not available cross jobs

ovidiub13:


src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
```diff
@@ -1,6 +1,7 @@
 <div>
     Saves a set of files for use later in the same build, generally on another node/workspace.
```
dwnusbaum: I think the fact that stashes are not available across jobs is already covered by the phrasing `same build` here. Maybe we should bold `same build`?

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
```diff
@@ -1,6 +1,7 @@
...
     Stashed files are not otherwise available and are generally discarded at the end of the build.
+    Stashes are not available to be used cross jobs. For this you need to use <code>archiveArtifacts</code>/<code>copyArtifacts</code>.
```
dwnusbaum: `copyArtifacts` comes from a non-core plugin, so I don't think we want to recommend it directly. Maybe tweak the wording to something like this:
```suggestion
    If you want artifacts to persist after the build, consider using <a href="https://jenkins.io/doc/pipeline/steps/core/#-archiveartifacts-%20archive%20the%20artifacts"><code>archiveArtifacts</code></a> instead.
```

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
ovidiub13: AFAIK the Jenkins terminology is:

Jenkins has multiple `jobs`, that have multiple `builds`.
Each `build` can be a `pipeline` that has multiple `stages`.
Each `stage` has multiple `steps`.

So using the term `build` here is misleading.

From a build (pipeline) a job can trigger another job. The statement `on another node/workspace` may mislead users to believe that they can stash between the triggered job and the trigger job.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
ovidiub13: I like this variant more. What do you think?
```suggestion
    Stashes are not available to be used cross jobs. If you want artifacts to persist after the build, consider using <a href="https://jenkins.io/doc/pipeline/steps/core/#-archiveartifacts-%20archive%20the%20artifacts"><code>archiveArtifacts</code></a> instead.
```

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
dwnusbaum: Yes, that is the terminology, `same build` means that the stash can only be used from the same build of the same job. You cannot access a stash from a different build of the same job, or from a build of a different job entirely. I think `on another node/workspace` is mentioned because if you are still in the same node block, then you can just use a shell script to access the files directly, so there is no point of using `stash`. Maybe changing `same build` to `same build of the same job` to be more explicit would help?

ovidiub13: You are right, the phrasing is clear. But for someone who doesn't work with Jenkins all day every day, a more explicit phrasing could make the difference for hours of frustration. I don't think we need to change this line at all. The bellow completion clarifies if someone had any doubts.

bitwiseman: @ovidiub13 
It might be worth changing this line to 
```suggestion
    Saves a set of files for use later in the same Pipeline run, generally on another node/workspace.
```

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
bitwiseman: Compromise:
```suggestion
    Stashes from a particular pipeline run are not available in other runs or other pipelines.
    If you want artifacts to persist after the build, consider using 
    <a href="https://jenkins.io/doc/pipeline/steps/core/#archiveartifacts-archive-the-artifacts"
    <code>archiveArtifacts</code></a> instead.
```

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
bitwiseman: For pipelines this is more accurate:
```suggestion
    Saves a set of files for use later on any node/workspace in the same Pipeline run.
```

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
```diff
@@ -1,6 +1,7 @@
 <div>
     Saves a set of files for use later in the same build, generally on another node/workspace.
     Stashed files are not otherwise available and are generally discarded at the end of the build.
```
bitwiseman: This is a test comment.

dwnusbaum: "preserveStashes" is a Declarative thing, it doesn't exist as part of this plugin: https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/7502d14aa4db891302abf48a2c7de7c9ced97f95/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java#L53.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
ovidiub13: This still can lead to the misconception that a triggered build is part of this pipeline, or maybe that's just me... Anyway, I like this last suggestion the best. Thank you both.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help.html
ovidiub13: I'm ok with @bitwiseman 's suggestion, and I'd commit it, but I'm not sure how much weigh does @dwnusbaum counter argument has. I'll wait for you guys to settle this.

bitwiseman: @dwnusbaum 
I see your point,  but I felt we needed to mention it, otherwise someone using this step as part of Declarative Pipeline might either assume they had to restash files or might not know how to preserve stashes are across run restarts.   How about this? 

```suggestion
    By default, stashed files are discarded at the end of a pipeline run.
    Other plugins may change this behavior to preserve stashes for longer.
    For example, Declarative Pipeline includes a <code>preserveStashes()</code> option
    to allow stashes from a run to be retain and used if that run is restarted. 
    <br>
```

dwnusbaum: Yeah, this is the classic Jenkins problem of plugins not mapping well to holistic user-facing features. As long as you specify clearly that the option is part of Declarative, like in your latest suggestion, it seems fine to mention it.

bitwiseman: @ovidiub13
Okay, it sounds like that second suggestion looks good to dwnusbaum.

ovidiub13: Awesome. Please correct `from a run to be retain and` into `from a run to be retained and` and I'll commit it. I can't seem to change it, and I want to give you credit for that change and keep it your commit.

bitwiseman: @ovidiub13 Done.

PR #90, opened 2019-07-03
[JENKINS-58322] feat: IgnoreError step

kuisathaverat:
see [JENKINS-58322](https://issues.jenkins-ci.org/browse/JENKINS-58322)

On the issue [JENKINS-45579](https://issues.jenkins-ci.org/browse/JENKINS-45579) @dwnusbaum has added a couple of really nice steps `WarnErrorStep`and `CatchErrorStep`, this improvement is a continuation of his work (mostly copy so I've maintained the same license header) to add another user case, when you need to ignore the failure of a stage but you want to mar the build as successful and it is not possible to use `CatchErrorStep` or you want to simplify your stage code, `ignoreErrors` option would mark the build as `SUCCESS` and the stage as `UNSTABLE`.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/IgnoreErrorStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/IgnoreErrorStep.java
@@ -1,0 +1,121 @@
+/*
+ * The MIT License
+ *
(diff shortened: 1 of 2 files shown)


kuisathaverat:
Devin point me in the right direction, this PR is not useful

PR #95, opened 2019-07-23
Using a BOM

jglick:
@oleg-nenashev this case is of interest since it is applied to a plugin which is itself mentioned in the BOM. See https://github.com/apache/maven-integration-testing/pull/25.


pom.xml
```diff
@@ -201,7 +187,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>${workflow-api-plugin.version}</version>
+            <version>2.35</version> <!-- TODO https://github.com/jenkinsci/bom/pull/32 -->
```
jglick: https://github.com/jenkinsci/bom/pull/34 to be precise.

dwnusbaum:
Are there any problems with using the BOM in the context of security releases or when working with incremental/snapshot versions of dependencies? Or would you just have to temporarily add an explicit `<version>`, add the new version to the BOM once it gets released, and then update the version of the BOM used by the plugin and remove the explicit uses of `<version>`?

jglick:
> would you just have to…

Yes.

jglick:
On hold pending https://github.com/jenkinsci/bom/pull/107.

pom.xml
```diff
@@ -28,15 +28,15 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.43</version>
+        <version>3.50</version>
```
bitwiseman:
```suggestion
        <version>3.56</version>
```

PR #97, opened 2019-09-22
Added delay capability to retry

krotte1:
In some cases a user wants to add a delay between retries that are performed. While continue to support the original usage of retry, this adds three new attributes to RetryStep: useTimeDelay, timeDelay, and units. If useTimeDelay is true, then the timeDelay will be applied between retries.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStep.java
@@ -26,26 +26,62 @@ package org.jenkinsci.plugins.workflow.steps;
 
 import hudson.Extension;
 import hudson.model.TaskListener;
+import hudson.util.ListBoxModel;
+
+import java.io.Serializable;
 import java.util.Collections;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * Executes the body up to N times.
  *
  * @author Kohsuke Kawaguchi
  */
-public class RetryStep extends Step {
+public class RetryStep extends Step implements Serializable {
     
     private final int count;
+    private int timeDelay;
+    private TimeUnit unit = TimeUnit.SECONDS;
+    private boolean useTimeDelay = false;
+    
+    public int left;
 
     @DataBoundConstructor
     public RetryStep(int count) {
         this.count = count;
+        this.left = count;
     }
 
     public int getCount() {
         return count;
+    }
+
+    @DataBoundSetter public void setUseTimeDelay(boolean useTimeDelay) {
+        this.useTimeDelay = useTimeDelay;
+    }
+
+    public boolean isUseTimeDelay() {
+        return useTimeDelay;
+    }
+
+    @DataBoundSetter public void setTimeDelay(int timeDelay) {
+        this.timeDelay = timeDelay;
+    }
+
+    public int getTimeDelay() {
+        return timeDelay;
+    }
+
+    @DataBoundSetter public void setUnit(TimeUnit unit) {
+        this.unit = unit;
+    }
+
+    public TimeUnit getUnit() {
+        return unit;
     }
 
     @Override
@@ -55,7 +91,7 @@ public class RetryStep extends Step {
 
     @Override
     public StepExecution start(StepContext context) throws Exception {
-        return new RetryStepExecution(count, context);
+        return new RetryStepExecution(this, context);
     }
 
     @Extension
@@ -76,6 +112,14 @@ public class RetryStep extends Step {
             return "Retry the body up to N times";
         }
 
+        public ListBoxModel doFillUnitItems() {
+            ListBoxModel r = new ListBoxModel();
+            for (TimeUnit unit : TimeUnit.values()) {
+                r.add(unit.name());
+            }
+            return r;
+        }
+        
         @Override
         public Set<? extends Class<?>> getRequiredContext() {
             return Collections.singleton(TaskListener.class);
@@ -83,4 +127,5 @@ public class RetryStep extends Step {
 
     }
 
+    private static final long serialVersionUID = 1L;
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
@@ -3,49 +3,150 @@ package org.jenkinsci.plugins.workflow.steps;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.AbortException;
 import hudson.Functions;
+import hudson.Util;
 import hudson.model.TaskListener;
+
+import java.util.UUID;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+
+import com.google.common.base.Function;
+
+import jenkins.util.Timer;
 
 /**
  * @author Kohsuke Kawaguchi
  */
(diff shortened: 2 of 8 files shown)


krotte1:
@jglick  @basil  @dwnusbaum I figured it might be worth reaching out to a few of you with recent activity on the workflow-basic-steps-plugin. I have added the ability to introduce a delay in between retries and would like you to review this pull-request. Thank you

basil:
Hey @krotte1, unlike Jesse and Devin I am not (yet!) an official committer to this project, so I defer to them for any official feedback. My comments below are merely suggestions, not blocking issues.

The feature seems useful, and I've wanted it in my own pipelines several times. I haven't reviewed the implementation yet, but I took a look at the proposed API change to the `retry` step and I have some ideas about that. Your proposed syntax is as follows:

```groovy
retry(count: 3, timeDelay: 10, unit: 'SECONDS', useTimeDelay: true) {
  [...]
}
```

Whenever I'm adding a new API, I try to study existing examples of similar APIs to learn from their mistakes and possibly gain insights that could benefit the design of my own API. In this case, I took a look at the API offered by [Tenacity](https://tenacity.readthedocs.io/en/latest/#waiting-before-retrying), which is (to me) the gold standard in retry APIs. I use it frequently in my Python-based build scripts.

Tenacity offers several options when it comes to waiting before retries. I've used several of them myself, depending on the use case. Some examples:

- Fixed period: `@retry(wait=wait_fixed(2))`
- Randomness: `@retry(wait=wait_random(min=1, max=2))`
- Exponential backoff: `@retry(wait=wait_exponential(multiplier=1, min=4, max=10))`
- Fixed waits and jitter: `@retry(wait=wait_fixed(3) + wait_random(0, 2))`
- Exponentially increasing jitter: `@retry(wait=wait_random_exponential(multiplier=1, max=60))`
- Chain of backoffs: `@retry(wait=wait_chain(*[wait_fixed(3) for i in range(3)] + [wait_fixed(7) for i in range(2)] + [wait_fixed(9)]))`

As you can see, Tenacity is the "Cadillac" of retry APIs. Am I suggesting that you implement all of this in your PR? I am emphatically _not_ suggesting that.

What I am suggesting, though, is that you [design for extension](https://checkstyle.sourceforge.io/config_design.html#DesignForExtension) so that these additional features could be added later.

For example, it might be worth sketching out some possible syntax for the fancier forms of time-based retry (such as backoff and jitter) and making sure that those future improvements could be made without having to break compatibility with the syntax being proposed in this change. That way, we can move forward with what you're proposing here with more confidence that future extensions won't run into implementation issues.

What do you think?

krotte1:
Thanks @basil. I like where you are going with this and can see the usefulness of the implementation you suggested. Before I go through the effort of refactoring the code to make this type of implementation possible, I would like to know from @jglick @dwnusbaum @abayer would even allow this enhancement or the more extensible implementation you propose into the baseline.

krotte1:
I was really hoping to hear from one of the maintainers (@jglick @dwnusbaum @kohsuke @abayer) about the possibility of incorporating this time of feature in the baseline. I have started implementing what @basil talked about on a different branch and am almost finished.... I don't want to continue down this line if the simple case won't even be accepted.

krotte1:
I now have the extensible version implemented that was mentioned by @basil in a previous post.  @jglick @dwnusbaum @kohsuke @abayer ... I was just wondering if you could tell me if either of these implementations would be accepted. The simple version in this pull request justs adds a fixed delay option to the retry step. The extensible version that I could merge into the pull-request allows for a FixedDelay, RandomDelay, IncrementalDelay, ExponentialDelay, and a RandomExponential delay. It can also be easily extended to add other delay algorithms.

krotte1:
I looked around and see that some pull requests have reviewers added to them, but I can't figure out how to do it myself. I beginning to think that only folks like @jglick @dwnusbaum @kohsuke @abayer that may have some type of elevated privileges can do it. All of the normal interactions with the Reviewers tab are not available to me so I can't really request someone to review this.

jglick:
Indeed you need write permission to a repository to modify reviewers. There is a new “triage” permission on GitHub that we are proposing to roll out; see the Jenkins dev list for more.

krotte1:
@jglick - thank you for adding the reviewers to this pull request. I really appreciate it..

@basil - I went ahead and implemented the recommendations you made earlier in this pull request. Do you want me to go ahead and merge that into this pull request before you review it? I made the delay extensible and implemented many of the items you mentioned above: FixedDelay, RandomDelay, IncrementalDelay, ExponentialDelay, and a RandomExponential.

dwnusbaum:
@krotte1 we don't currently have a great process for evaluating new features like this, but we are working on improving it. For now, can you [create a Jira ticket](https://issues.jenkins-ci.org), component `workflow-basic-steps`, mention me in the description (`@dnusbaum`), and we can discuss the idea/design there?

krotte1:
A new Jira ticket (https://issues.jenkins-ci.org/browse/JENKINS-59678) was created for this pull request to discuss the desired solution (simple fixed retry in the pull request or extensible solution that has already been implemented and can be added to this pull request).

krotte1:
@dwnusbaum - now that https://issues.jenkins-ci.org/browse/JENKINS-59678 has been out there for a week, what happens next? Which solution do we want to go with - simple fixed time delay between retries or an extensible method with different algorithms for retry delays?

krotte1:
@dwnusbaum - just wondering what the status is... I am really hoping to get one of these delay solutions incorporated into the baseline.

src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStep.java
```diff
@@ -26,36 +26,72 @@
...
     public RetryStep(int count) {
         this.count = count;
+        this.left = count;
     }
 
     public int getCount() {
         return count;
     }
 
+    @DataBoundSetter public void setUseTimeDelay(boolean useTimeDelay) {
```
timja: Is this needed? shouldn't a non null / non default time delay mean this is enabled?

src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStep.java
```diff
@@ -76,11 +112,20 @@ public String getDisplayName() {
...
+        }
+        
         @Override
         public Set<? extends Class<?>> getRequiredContext() {
             return Collections.singleton(TaskListener.class);
         }
 
     }
 
+    private static final long serialVersionUID = 1L;
```
timja: normally constants are at the top of a class in java

src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
```diff
@@ -1,51 +1,153 @@
...
+
+    RetryStepExecution(@Nonnull RetryStep step, StepContext context) {
+        super(context);
+        this.step = step;
+        this.count = step.getCount();
     }
 
-    @Override
-    public boolean start() throws Exception {
+    @Override public boolean start() throws Exception {
```
timja: `@Override` would normally go above a method, and it already was above there before you changed it so I would suggest you move it back

timja: `@Override` would normally go above a method, and it already was above there before you changed it so I would suggest you move it back

v1v:
Awesome! Great work!

I just found this PR, as I actually had to create a few retry/sleep steps and came up with some internal shared library step that it works but I wish we could have this feature in place.

Any plans to move forward this PR?

Thanks

jakauppila:
Is there anything holding up this PR from being merged?

src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStep.java
car-roll: It does seem to be a bit extra here

PR #98, opened 2019-09-26
[JENKINS-27127] Proof that waitUntil can be used to implement a semaphore

jglick:
Just here as a demonstration of one use case of [JENKINS-27127](https://issues.jenkins-ci.org/browse/JENKINS-27127); I do not think it needs to be merged.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStepTest.java
@@ -40,6 +40,7 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runners.model.Statement;
 import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStepTest.java
```diff
@@ -194,6 +195,32 @@
...
+        story.then(r -> {
+            WorkflowJob p = r.createProject(WorkflowJob.class);
+            p.setDefinition(new CpsFlowDefinition(
+                "def semaphore = 10\n" +
+                "def rand = new Random()\n" +
+                "def branches = [:]\n" +
+                "for (int i = 0; i < 50; i++) {\n" +
+                "  def branch = /b$i/\n" +
+                "  branches[branch] = {\n" +
+                "    waitUntil {if (semaphore > 0) {semaphore--; true} else {false}}\n" +
```
jglick: Since the CPS VM uses green threads, and the closure uses no continuations, it should run atomically. Therefore there is no need for lower-level Java synchronization primitives.

PR #101, opened 2019-11-19
ErrorStep to write error message to console output

ArturHarasimiuk:
After this change it is possible to get error message from 'Pipeline Steps' view in log view. It is very useful as it allows to check error messages while build is still running.

Signed-off-by: Artur Harasimiuk <artur.harasimiuk@intel.com>

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/ErrorStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/ErrorStep.java
@@ -27,6 +27,7 @@ package org.jenkinsci.plugins.workflow.steps;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.AbortException;
 import hudson.Extension;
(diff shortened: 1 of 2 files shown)


dwnusbaum:
Thanks for the PR!

> It is very useful as it allows to check error messages while build is still running.

This doesn't make sense to me. The point of this step is that is causes the build to fail immediately by throwing an exception. The message for the exception will be printed to the overall build log as long as your Pipeline does not do something to catch the exception.

If you are using `try`/`catch` in your Pipelines to catch the exception thrown by this step, and then suppressing the exception, I think it would make more sense to instead log the message in the `catch` block, which should result in similar behavior.

Maybe what you want is something more like the `unstable` step that changes the build result and marks the step with a non-successful result, but that doesn't actually throw an exception?

dwnusbaum:
I'm going to close this PR for now, but feel free to respond to my comment and reopen it if you still want to work on it.

PR #105, opened 2020-01-27
[JENKINS-46124] - support map with EnvStep

jetersen:
fixes [JENKINS-46124](https://issues.jenkins-ci.org/browse/JENKINS-46124)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
@@ -178,7 +178,7 @@ public class EnvStep extends Step {
 
         private static List<String> toKeyValueList(Map<?, ?> map) {
             return map.entrySet().stream()
-                .map(m -> m.getKey() + "=" + (m.getValue() == null ? "" : m.getValue().toString()))
+                .map(m -> (String) m.getKey() + "=" + (m.getValue() == null ? "" : m.getValue().toString()))
                 .collect(Collectors.toList());
         }
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.steps;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.EnvVars;
 import hudson.Extension;
@@ -34,7 +35,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import net.sf.json.JSONObject;
+import org.jenkinsci.plugins.structs.describable.CustomDescribableModel;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.StaplerRequest;
 
@@ -103,7 +106,7 @@ public class EnvStep extends Step {
         }
     }
 
-    @Extension public static class DescriptorImpl extends StepDescriptor {
+    @Extension public static class DescriptorImpl extends StepDescriptor implements CustomDescribableModel {
 
         @Override public String getFunctionName() {
             return "withEnv";
@@ -155,6 +158,25 @@ public class EnvStep extends Step {
             }
         }
 
+        @NonNull
+        @Override
+        public Map<String, Object> customInstantiate(@NonNull Map<String, Object> arguments) {
+            Map<String, Object> r = new HashMap<>(arguments);
+            Object overrides = r.get("overrides");
+            if (overrides instanceof Map) {
+                r.put("overrides", toKeyValueList((Map<?, ?>) overrides));
+            }
+            return r;
+        }
+
+        private static List<String> toKeyValueList(Map<?, ?> map) {
+            return map.entrySet().stream()
+                .filter(e -> e.getKey() instanceof String)
+                .collect(Collectors.toMap(e -> (String)e.getKey(), e -> e.getValue() == null ? "" : (String) e.getValue()))
+                .entrySet().stream()
+                .map(m -> m.getKey() + "=" + m.getValue())
+                .collect(Collectors.toList());
+        }
     }
 
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
@@ -178,10 +178,7 @@ public class EnvStep extends Step {
 
         private static List<String> toKeyValueList(Map<?, ?> map) {
             return map.entrySet().stream()
-                .filter(e -> e.getKey() instanceof String)
-                .collect(Collectors.toMap(e -> (String)e.getKey(), e -> e.getValue() == null ? "" : e.getValue().toString()))
-                .entrySet().stream()
-                .map(m -> m.getKey() + "=" + m.getValue())
+                .map(m -> m.getKey() + "=" + (m.getValue() == null ? "" : m.getValue().toString()))
                 .collect(Collectors.toList());
         }
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
@@ -153,6 +153,12 @@ public class EnvStep extends Step {
                     }
                 }
                 return b.toString();
+            } else if (overrides instanceof Map) {
+                return ((Map<?, ?>) overrides).keySet()
+                    .stream()
+                    .filter(e -> e instanceof String)
+                    .map(Object::toString)
+                    .collect(Collectors.joining(", "));
             } else {
                 return null;
             }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
@@ -178,7 +178,7 @@ public class EnvStep extends Step {
         private static List<String> toKeyValueList(Map<?, ?> map) {
             return map.entrySet().stream()
                 .filter(e -> e.getKey() instanceof String)
-                .collect(Collectors.toMap(e -> (String)e.getKey(), e -> e.getValue() == null ? "" : (String) e.getValue()))
+                .collect(Collectors.toMap(e -> (String)e.getKey(), e -> e.getValue() == null ? "" : e.getValue().toString()))
                 .entrySet().stream()
                 .map(m -> m.getKey() + "=" + m.getValue())
                 .collect(Collectors.toList());
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
@@ -168,9 +168,10 @@ public class EnvStep extends Step {
         @Override
         public Map<String, Object> customInstantiate(@NonNull Map<String, Object> arguments) {
             Map<String, Object> r = new HashMap<>(arguments);
-            Object overrides = r.get("overrides");
+            final String key = "overrides";
+            Object overrides = r.get(key);
             if (overrides instanceof Map) {
-                r.put("overrides", toKeyValueList((Map<?, ?>) overrides));
+                r.put(key, toKeyValueList((Map<?, ?>) overrides));
             }
             return r;
         }
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
@@ -245,6 +245,26 @@ public class EnvStepTest {
         });
     }
 
+    @Test public void mapNumbersNested() {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(
+                    "node {\n" +
+                        "  withEnv([A: 1]) {\n" +
+                        "    withEnv([B: 2]) {\n" +
+                        "      withEnv([C: true]) {\n" +
+                        "        isUnix() ? sh('echo A=$A B=$B C=$C') : bat('echo A=%A% B=%B% C=%C%')\n" +
+                        "      }\n" +
+                        "    }\n" +
+                        "  }\n" +
+                        "}", true));
+                WorkflowRun b = story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
+                story.j.assertLogContains("A=1 B=2 C=true", b);
+            }
+        });
+    }
+
     @Test public void mapNested() {
         story.addStep(new Statement() {
             @Override public void evaluate() throws Throwable {
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
@@ -83,36 +83,6 @@ public class EnvStepTest {
         });
     }
 
-    @Test public void mapOverriding() {
-        story.addStep(new Statement() {
-            @Override public void evaluate() throws Throwable {
-                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
-                p.setDefinition(new CpsFlowDefinition(
-                    "env.CUSTOM = 'initial'\n" +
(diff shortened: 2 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
```diff
@@ -83,6 +83,36 @@
...
+                story.j.assertLogContains(Functions.isWindows() ? "inside CUSTOM=override NOVEL=val BUILD_TAG=custom NULLED= FOOPATH=C:\\ball;C:\\foos;" : "inside CUSTOM=override NOVEL=val BUILD_TAG=custom NULLED= FOOPATH=/opt/ball:/opt/foos:", b);
+                story.j.assertLogContains("groovy NULLED=null", b);
+                story.j.assertLogContains("outside CUSTOM=initial NOVEL= NULLED=outside", b);
+                List<FlowNode> coreStepNodes = new DepthFirstScanner().filteredNodes(
+                    b.getExecution(),
+                    Predicates.and(
+                        new NodeStepTypePredicate("withEnv"),
+                        n -> n instanceof StepStartNode && !((StepStartNode) n).isBody()));
+                assertThat(coreStepNodes, Matchers.hasSize(1));
+                assertEquals("CUSTOM, NOVEL, BUILD_TAG, NULLED, FOOPATH+BALL", ArgumentsAction.getStepArgumentsAsString(coreStepNodes.get(0)));
```
casz: This test case fails, I am not sure why it seems to be doing the right thing but unable to get the step arguments.

casz: d0h I had to write the converter for arguments to string.

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
```diff
@@ -150,11 +153,36 @@
...
             } else {
                 return null;
             }
         }
 
+        @NonNull
+        @Override
+        public Map<String, Object> customInstantiate(@NonNull Map<String, Object> arguments) {
+            Map<String, Object> r = new HashMap<>(arguments);
+            Object overrides = r.get("overrides");
```
MindaugasLaganeckas: DRY for "overrides" :)

MindaugasLaganeckas (approved):
A nice feature and a nice set of tests! :)

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
```diff
@@ -150,11 +153,37 @@
                     }
                 }
                 return b.toString();
+            } else if (overrides instanceof Map) {
```
jglick: BTW I lobbied for `ArgumentsAction[Impl]` to persist only the internal, resolved form of arguments, but the design we went with persists the surface form, necessitating extra work like this.

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
```diff
@@ -150,11 +153,37 @@
...
+            Object overrides = r.get(key);
+            if (overrides instanceof Map) {
+                r.put(key, toKeyValueList((Map<?, ?>) overrides));
+            }
+            return r;
+        }
+
+        private static List<String> toKeyValueList(Map<?, ?> map) {
+            return map.entrySet().stream()
+                .filter(e -> e.getKey() instanceof String)
```
jglick: Clearer IMO to just cast to `String`, so that people will get a clear error if they use something else.
```suggestion
```

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
```diff
@@ -150,11 +153,37 @@
...
+                .collect(Collectors.toMap(e -> (String)e.getKey(), e -> e.getValue() == null ? "" : e.getValue().toString()))
+                .entrySet().stream()
+                .map(m -> m.getKey() + "=" + m.getValue())
```
jglick: Why are you collecting a map and then restreaming it? Would seem to simpler to just

```java
.map(m -> (String) e.getKey() + "=" + (m.getValue() == null ? "" : m.getValue()))

File: src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java [hunk 6]
```diff
@@ -83,6 +83,36 @@
         });
     }
 
+    @Test public void mapOverriding() {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(
+                    "env.CUSTOM = 'initial'\n" +
+                        "env.FOOPATH = node {isUnix() ? '/opt/foos' : 'C:\\\\foos'}\n" +
```
jglick: You need not copy all these exotic corner cases from overriding. Should suffice to have a simple script like

```groovy
withEnv([a: 1, b: 2]) {
  echo "a=$a b=$b"
}
```

src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
```diff
@@ -105,6 +135,28 @@
         });
     }
 
+    @Test public void mapParallel() {
```
jglick: Again, unnecessary, adequately covered by existing tests; all that is changing is the syntax.

src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
```diff
@@ -140,6 +192,41 @@
         });
     }
 
+    @Test public void mapRestarting() {
```
jglick: Ditto.

src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
```diff
@@ -158,6 +245,44 @@
         });
     }
 
+    @Test public void mapNumbersNested() {
```
jglick: Some simpler variant of this may still be useful, to assert that map syntax allows non-string values.

src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
```diff
@@ -158,6 +245,44 @@
...
+                        "    }\n" +
+                        "  }\n" +
+                        "}", true));
+                WorkflowRun b = story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
+                story.j.assertLogContains("A=1 B=2 C=true", b);
+            }
+        });
+    }
+
+    @Test public void mapNested() {
```
jglick: Probably can be skipped.

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
```diff
@@ -150,11 +153,37 @@
                     }
                 }
                 return b.toString();
+            } else if (overrides instanceof Map) {
+                return ((Map<?, ?>) overrides).keySet()
+                    .stream()
+                    .filter(e -> e instanceof String)
+                    .map(Object::toString)
+                    .collect(Collectors.joining(", "));
```
jglick: Not covered by tests AFAICT?

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
```diff
@@ -150,11 +153,37 @@
...
             } else {
                 return null;
             }
         }
 
+        @NonNull
+        @Override
+        public Map<String, Object> customInstantiate(@NonNull Map<String, Object> arguments) {
+            Map<String, Object> r = new HashMap<>(arguments);
+            final String key = "overrides";
```
jglick: This design forces people to write, say,

```groovy
withEnv([VAR: 'value']) {/* … */}
```

rather than the possibly more natural

```groovy
withEnv(VAR: 'value') {/* … */}
```

which would be available for anything other than an environment variable named (weirdly) `overrides`.

jglick (commented):
Do you want to implement `customUninstantiate` to guide users of **Pipeline Syntax** to use the new form, or leave them with the `List<String>` syntax?

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
casz: Definitely covered by tests as it was failing before I created the converter 😆

jglick: Meaning tests verify that this method returns normally when a map is used, not that the return value is correct.

casz: No there is a test that looks at the output.
https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/105/files#diff-c2e520244ad09761e2187513ef5c751cR111

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
casz: Hmmm I think we need to support both as we are trying to support passing a map from scm checkout

casz: it's named overrides because that's how groovy passes the object.
image

casz: Even when I pass it like you suggested
```java
withEnv(VAR: 'value') {/* … */}
```
I still get `arguments.overrides` with whatever arguments inside.

casz: I'll add a test case for this

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
casz: To the two tests I added now I am testing the output of the function.

casz:
please take another look @jglick I think I listen to your feedback, I honestly don't know what I was thinking about when I wrote that stream collector 🤯

src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
```diff
@@ -245,40 +158,48 @@
...
                 WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
                 p.setDefinition(new CpsFlowDefinition(
                     "node {\n" +
-                        "  withEnv([A: 1]) {\n" +
-                        "    withEnv([B: 2]) {\n" +
-                        "      withEnv([C: true]) {\n" +
-                        "        isUnix() ? sh('echo A=$A B=$B C=$C') : bat('echo A=%A% B=%B% C=%C%')\n" +
-                        "      }\n" +
-                        "    }\n" +
+                        "  withEnv(a: 1, b: 2, c: 'hello world', d: true) {\n" +
```
jglick: test null values

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
```diff
@@ -178,10 +178,7 @@
 
         private static List<String> toKeyValueList(Map<?, ?> map) {
             return map.entrySet().stream()
-                .filter(e -> e.getKey() instanceof String)
-                .collect(Collectors.toMap(e -> (String)e.getKey(), e -> e.getValue() == null ? "" : e.getValue().toString()))
-                .entrySet().stream()
-                .map(m -> m.getKey() + "=" + m.getValue())
+                .map(m -> m.getKey() + "=" + (m.getValue() == null ? "" : m.getValue().toString()))
```
jglick:
```suggestion
                .map(m -> (String) m.getKey() + "=" + (m.getValue() == null ? "" : m.getValue().toString()))
```

since non-`String` keys would be bogus

src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
```diff
@@ -245,40 +158,48 @@
...
+                        "    echo \"a=$a b=$b c=$c d=$d\"" +
```
jglick: arguably more legibly

```suggestion
                        "    echo(/a=$a b=$b c=$c d=$d/)" +
```

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
casz: IDE consider this a redundant cast and wants to remove it.

src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
casz: the customInstate does the right with with `e: null` turning into `e=` however it throws a `groovy.lang.MissingPropertyException: No such property: e for class: groovy.lang.Binding`

casz: ah when null we have to echo like this `E=${env.E}`

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
jglick: Then your IDE is wrong and you should ignore it. You should be able to have a test demonstrating a failure from some bogus thing like

```groovy
withEnv([new Object(): 'x']) {}
```

casz: This case is not supported:
```
WorkflowScript: 2: illegal colon after argument expression;
  solution: a complex label expression before a colon must be parenthesized @ line 2, column 72.
  D: true, E: null, new Object(): 'x']) {
```

jglick: Maybe

```groovy
withEnv([(new Object()): 'x']) {}
```

then, whatever the syntax is. At worst

```groovy
def m = [:]
m.put(new Object(), 'x')
withEnv(m) {}
```

casz: Than you get into the whole script security: one you add `(new Object())` `org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use new java.lang.Object`

casz: I'd argue your asking for trouble

jglick: OK forget `new Object()`, just some arbitrary object key which is _not_ a `String`. `new URL('http://nowhere.net/')` for example is whitelisted.

casz: Than you get into `java.lang.ClassCastException: java.net.URL cannot be cast to java.lang.String` 🤔 

Properly better to call `toString` instead of trying to cast the object

jglick: > Then you get into `java.lang.ClassCastException`

Exactly what I was requesting with 8d385a2.

casz: image

casz:
@bitwiseman @jglick 🙏 Would be a great addition to get merged.

dwnusbaum (changes requested):
Thanks for the PR! The implementation looks good to me (thanks for adding tests!). I am a little hesitant to add complexity since we will have to support both syntaxes going forward, but I think the trade-off is worth it in this case.

Can you please update [EnvStep/help-overrides.html](https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/f5586e7f38df6ab42688c3c968943874d418461a/src/main/resources/org/jenkinsci/plugins/workflow/steps/EnvStep/help-overrides.html) (and maybe also [EnvStep/help.html](https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/f5586e7f38df6ab42688c3c968943874d418461a/src/main/resources/org/jenkinsci/plugins/workflow/steps/EnvStep/help.html)) to mention the new syntax and give an example? Once that is done I am happy to merge the PR.

src/main/java/org/jenkinsci/plugins/workflow/steps/EnvStep.java
bitwiseman: @jglick  Okay, is there some action to be taken about that now?

jglick: Not easily. Just noting some background.

casz:
I'll see what I can do :)
However somewhat busy in this week. Starting at a new workplace 😁

mss:
@dwnusbaum I was looking for this exact feature for years and finally found this PR but it got stalled. I'd be happy to provide the requested documentation but since this code does not apply in the meantime I wondered who would have to adapt it to the changed codebase?

PR #106, opened 2020-02-07
[JENKINS-61019] Tolerate null killer during initialization

jsoref:
Hit while debugging

https://issues.jenkins-ci.org/browse/JENKINS-61019

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
@@ -136,7 +136,9 @@ public class TimeoutStepExecution extends AbstractStepExecutionImpl {
             }, delay, TimeUnit.MILLISECONDS);
         } else {
             listener().getLogger().println("Timeout expired " + Util.getTimeSpanString(- delay) + " ago");
-            cancel();
+            if (killer != null) {
+                cancel();
+            }
         }
     }
 


src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -136,7 +136,9 @@ public void run() {
             }, delay, TimeUnit.MILLISECONDS);
         } else {
             listener().getLogger().println("Timeout expired " + Util.getTimeSpanString(- delay) + " ago");
-            cancel();
+            if (killer != null) {
+                cancel();
+            }
```
dwnusbaum: So IIUC, what happened is that after the restart, `killer` was null since it is a transient field, and the body was supposed to have completed some time in the past, so the step tried to cancel the body, but then the condition in `cancel` that tries to prevent the body from being double-cancelled caused the NPE. With this change, I think the body would continue running with no timeout, right?

We could fix that by adding something like this:

```suggestion
            if (killer != null) {
                cancel();
            } else {
                // Resuming from a restart, body expired while the Pipeline was suspended.
                killer = Timer.get().submit(new Runnable() {
                    @Override
                    public void run() {
                        cancel();
                    }
                });
            }
```
I guess this opens up the question too of whether the timeout should respect the wall-clock time (as in my suggestion), or should only track time the Pipeline was actually running (not sure what would be needed to make that happen, and probably not something to try to handle in this PR). Maybe it is reasonable to just have things as in your patch where the timeout is just ignored in this case?

jsoref: I think the idea of allowing a pipeline to run unbounded when it was supposed to have a timeout is fairly wrong, which leads me to support your proposal.

Your proposal sounds reasonable. But I'd want someone to confirm... None of this is my area.

basil: >I guess this opens up the question too of whether the timeout should respect the wall-clock time (as in my suggestion), or should only track time the Pipeline was actually running (not sure what would be needed to make that happen, and probably not something to try to handle in this PR).

I have a master with a large number (thousands) of jobs that takes 30-90 minutes to start up after an upgrade. I personally like the behavior of counting only the time the Pipeline was running because it allows upgrades to take as long as they need to take without impacting running jobs.

P.S. I hit this bug myself during a recent upgrade and updated [JENKINS-61019](https://issues.jenkins-ci.org/browse/JENKINS-61019) with my stack trace.

dwnusbaum: > I personally like the behavior of counting only the time the Pipeline was running

Yeah, I think the question is just whether it is possible to implement that in a way that is robust without being too complex.

The other thing I am curious about, but don't have time to check at the moment, is whether this issue has always existed, or if it is a regression, in which case I want to know what the previous behavior was.

jsoref: I can't speak to this. I only hit it because I was actively futzing w/ a debugger (including leaving it stopped overnight)

dwnusbaum: I looked into this a bit more today and came up with a test case. It looks like it was originally reported as [JENKINS-42940](https://issues.jenkins-ci.org/browse/JENKINS-42940). The bug only happens when `TimeoutStepExecution.forcible` is already `true` when the step resumes, which I guess is why it's not a super common issue, and it was hard to get a test case to save `program.dat` at the right time to be able to reproduce the issue. I think the easiest fix that still causes the body to be cancelled is to make `forcible` transient, so that after a restart we try to cancel the body normally, which will set up everything correctly for a force-cancel after a minute if needed.

jsoref: @dwnusbaum: can i get you to write a fix for this? It's beyond my focus...

dwnusbaum: @jsoref Yeah, I will try to file a PR in the next few days, just wanted to document my findings in the meantime.

dwnusbaum:
Ok, I filed https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/112 with the fix and test case I mentioned [here](https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/106#discussion_r397466133), so I am going to close this PR.

Thanks for filing the PR @jsoref!

PR #107, opened 2020-02-19
JENKINS-60239 Aborted catchError sets build result to ABORTED if fail fast is true

daanipuui:
Hello,

I like strongly typed solutions but in this case the downside is that I have introduced a new dependency (workflow-cps) which might not be something you want or like.

I could pass a message to the `FailFastException` and check if any cause of interruption has that message (it's not very elegant and hides the dependency on the parallel step).

Daniel

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,10 +66,10 @@
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.150.1</jenkins.version>
         <java.level>8</java.level>
-        <workflow-step-api-plugin.version>2.22</workflow-step-api-plugin.version>
-        <workflow-cps-plugin.version>2.70</workflow-cps-plugin.version>
+        <workflow-step-api-plugin.version>2.23-SNAPSHOT</workflow-step-api-plugin.version>
+        <workflow-cps-plugin.version>2.81-SNAPSHOT</workflow-cps-plugin.version>
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
-        <workflow-api-plugin.version>2.34</workflow-api-plugin.version>
+        <workflow-api-plugin.version>2.36</workflow-api-plugin.version>
         <useBeta>true</useBeta>
         <git-plugin.version>4.0.0-beta3</git-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -66,10 +66,10 @@
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.150.1</jenkins.version>
         <java.level>8</java.level>
-        <workflow-step-api-plugin.version>2.22</workflow-step-api-plugin.version>
-        <workflow-cps-plugin.version>2.70</workflow-cps-plugin.version>
+        <workflow-step-api-plugin.version>2.23-SNAPSHOT</workflow-step-api-plugin.version>
+        <workflow-cps-plugin.version>2.81-SNAPSHOT</workflow-cps-plugin.version>
```
dwnusbaum: Just for future reference, if you have other PRs that are linked to a PR, please include a link to them in the PR description. Instead of using SNAPSHOT dependencies, you can use [incremental dependencies](https://jenkins.io/blog/2018/05/15/incremental-deployment/), for example [here](https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/workflow/workflow-step-api/2.23-rc561.8c6752584d39/) is the incremental build for https://github.com/jenkinsci/workflow-step-api-plugin/pull/55.

src/test/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStepTest.java
```diff
@@ -234,6 +234,22 @@
...
     }
 
+    @Issue("JENKINS-60239")
+    @Test public void parallelStageFails() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(""
+                + "node {\n"
+                + "    parallel(\n"
+                + "        stage1: { error 'fail' },\n"
+                + "        stage2: { catchError(stageResult: 'FAILURE', catchInterruptions: true) { sleep 10 } },\n"
```
dwnusbaum: I think you can make this change in your Pipeline to get the desired behavior:

```suggestion
                + "        stage2: { catchError(stageResult: 'FAILURE', catchInterruptions: false) { sleep 10 } },\n"
```

pom.xml
daanipuui: Hello @dwnusbaum , thanks for reviewing this and for your suggestion. I closed the issue because setting catchInterruptions to false works.
Thanks,
Best regards,

daanipuui:
catchInterruptions = false does the job

PR #108, opened 2020-02-26
Avoid NPE when calling fileExists step

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

fileExists can be called with a null parameter but will throw an NPE. Avoid the NPE by using a blank string instead.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/FileExistsStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/FileExistsStep.java
@@ -78,11 +78,9 @@ public final class FileExistsStep extends Step {
         }
 
         @Override protected Boolean run() throws Exception {
-            if (file == null) {
-                getContext().get(TaskListener.class).getLogger().println(Messages.FileExistsStep_NullString());
-                return getContext().get(FilePath.class).child("").exists();
-            } else if (StringUtils.isEmpty(file)) {
+            if (StringUtils.isEmpty(file)) {
                 getContext().get(TaskListener.class).getLogger().println(Messages.FileExistsStep_EmptyString());
+                return getContext().get(FilePath.class).child("").exists();
             }
         	return getContext().get(FilePath.class).child(file).exists();
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/FileExistsStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/FileExistsStep.java
@@ -78,7 +78,10 @@ public final class FileExistsStep extends Step {
         }
 
         @Override protected Boolean run() throws Exception {
-            if (StringUtils.isEmpty(file)) {
+            if (file == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/FileExistsStep.java
```diff
@@ -78,7 +78,10 @@ public String getFile() {
         }
 
         @Override protected Boolean run() throws Exception {
-            if (StringUtils.isEmpty(file)) {
```
bitwiseman: What about `StringUtils.IsBlank(file)`?  That also return `true` for null.  Do we need to log that the file was `null` rather than `""`?

src/main/resources/org/jenkinsci/plugins/workflow/steps/Messages.properties
```diff
@@ -1,4 +1,5 @@
 ArtifactArchiverStepExecution.Deprecated=The archive step is deprecated, please use archiveArtifacts instead.
 ArtifactArchiverStepExecution.NoFiles=No files found to archive for pattern "{0}"; continuing.
 ArtifactArchiverStepExecution.NoFilesWithExcludes=No files found to archive for pattern "{0}", excluding "{1}"; continuing.
-FileExistsStep.EmptyString=The fileExists step was called with the empty string, so the current directory will be checked instead.
+FileExistsStep.EmptyString=The fileExists step was called with an empty string, so the current directory will be checked instead.
```
bitwiseman:
```suggestion
FileExistsStep.EmptyString=The fileExists step was called with a null or empty string, so the current directory will be checked instead.
```

src/main/resources/org/jenkinsci/plugins/workflow/steps/Messages.properties
```diff
@@ -1,4 +1,5 @@
...
+FileExistsStep.NullString=The fileExists step was called with a null string, so the current directory will be checked instead.
```
bitwiseman:
```suggestion
```

src/main/java/org/jenkinsci/plugins/workflow/steps/FileExistsStep.java
res0nance: I guess the distinction doesn't matter. I'll fix this up. The important part is to not throw an NPE due to the return statement

bitwiseman: Oh, I see now.  How about unifying these paths instead?
```
file = StringUtils.isEmpty(file) ? "" : file;
```

bitwiseman: No NPE: absolutely. Thanks!

res0nance: I just changed the approach and called it with a blank string in the event its blank or null

PR #109, opened 2020-02-26
Add an IsMac step for convenience

res0nance:
This was manually tested with a macos node

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/IsMacStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/IsMacStep.java
@@ -3,6 +3,7 @@ package org.jenkinsci.plugins.workflow.steps;
 import hudson.Extension;
 import hudson.Launcher;
 import hudson.Platform;
+import hudson.remoting.VirtualChannel;
 import jenkins.security.MasterToSlaveCallable;
 
 import java.util.Collections;
@@ -20,14 +21,18 @@ public class IsMacStep extends Step {
         return new Execution(context);
     }
 
-    public static class Execution extends SynchronousStepExecution<Boolean> {
+    public static class Execution extends SynchronousNonBlockingStepExecution<Boolean> {
 
         Execution(StepContext context) {
             super(context);
         }
 
         @Override protected Boolean run() throws Exception {
-            return getContext().get(Launcher.class).getChannel().call(new IsMacOS());
+            VirtualChannel c = getContext().get(Launcher.class).getChannel();
+            if (c != null)
+                return c.call(new IsMacOS());
+            else
+                return false;
         }
 
         private static final long serialVersionUID = 1L;
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/IsMacStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/IsMacStep.java
@@ -1,0 +1,60 @@
+package org.jenkinsci.plugins.workflow.steps;
+
+import hudson.Extension;
+import hudson.Launcher;
+import hudson.Platform;
+import jenkins.security.MasterToSlaveCallable;
+
+import java.util.Collections;
+import java.util.Set;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+/**
+ * Checks whether we are running on Mac.
+ */
+public class IsMacStep extends Step {
+
+    @DataBoundConstructor public IsMacStep() {}
+
+    @Override public StepExecution start(StepContext context) throws Exception {
+        return new Execution(context);
+    }
+
+    public static class Execution extends SynchronousStepExecution<Boolean> {
+
+        Execution(StepContext context) {
+            super(context);
+        }
+
+        @Override protected Boolean run() throws Exception {
+            return getContext().get(Launcher.class).getChannel().call(new IsMacOS());
+        }
+
+        private static final long serialVersionUID = 1L;
+
+    }
+
+    @Extension public static final class DescriptorImpl extends StepDescriptor {
+
+        @Override public String getFunctionName() {
+            return "isMac";
+        }
+
+        @Override public String getDisplayName() {
+            return "Checks if running on a Mac node";
+        }
+
+        @Override public Set<? extends Class<?>> getRequiredContext() {
+            return Collections.singleton(Launcher.class);
+        }
+
+    }
+
+    private static final class IsMacOS extends MasterToSlaveCallable<Boolean, Exception> {
+        public Boolean call() {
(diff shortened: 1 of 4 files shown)


src/test/java/org/jenkinsci/plugins/workflow/steps/IsMacStepTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class IsMacStepTest {
+
+    @Rule public JenkinsRule r = new JenkinsRule();
+
+    @Test public void basics() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node {if (isMac()) sh 'sw_vers'}", true));
+        r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+    }
```
bitwiseman: So, this doesn't run  the `sh` unless the test is run on a mac.   Does `sw_vers` only exist on Mac? 

```suggestion
    @Test public void basics() throws Exception {
        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
        // Test passes on all platforms, on mac it will call "sw_vers", on other platforms it will not
        // "sw_vers" does not exist on unix or windows, so test would fail if "isMac()" were not working on those platforms. 
        p.setDefinition(new CpsFlowDefinition("node {if (isMac()) sh 'sw_vers'}", true));
        r.assertBuildStatusSuccess(p.scheduleBuild2(0));
    }
```

src/main/java/org/jenkinsci/plugins/workflow/steps/IsMacStep.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public class IsMacStep extends Step {
+
+    @DataBoundConstructor public IsMacStep() {}
+
+    @Override public StepExecution start(StepContext context) throws Exception {
+        return new Execution(context);
+    }
+
+    public static class Execution extends SynchronousNonBlockingStepExecution<Boolean> {
```
bitwiseman: This differs from

https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/ad0ec60819e306d9619768d2aecb72a144b61724/src/main/java/org/jenkinsci/plugins/workflow/steps/IsUnixStep.java#L44

Why?

res0nance: From the Javadoc of the `SynchronusStepExecution`
```
To call legacy Jenkins APIs which are potentially long-running and interruptible yet offer no asynchronous mode (for example because they block on a remoting call) use SynchronousNonBlockingStepExecution
```
I did it due to the remoting call.

src/main/java/org/jenkinsci/plugins/workflow/steps/IsMacStep.java
```diff
@@ -0,0 +1,65 @@
...
+
+        Execution(StepContext context) {
+            super(context);
+        }
+
+        @Override protected Boolean run() throws Exception {
+            VirtualChannel c = getContext().get(Launcher.class).getChannel();
+            if (c != null)
```
res0nance: I doubt this would be null but i have no idea what to do in cases where this is in fact null. Perhaps it should throw?

src/test/java/org/jenkinsci/plugins/workflow/steps/IsMacStepTest.java
res0nance: Yeah its only available in mac, I couldn't think of a better test

casz:
@res0nance not that I am against this step.
However this step does exist: https://github.com/jenkinsci/pipeline-utility-steps-plugin/blob/master/src/main/java/org/jenkinsci/plugins/pipeline/utility/steps/jenkins/NodesByLabelStep.java

src/main/java/org/jenkinsci/plugins/workflow/steps/IsMacStep.java
casz: c would be null when running on Jenkins master.
so it should definitely not throw as the Jenkins master could be running on mac.

src/main/java/org/jenkinsci/plugins/workflow/steps/IsMacStep.java
bitwiseman: @jglick  
Would this mean that `isUnixStep` should also be a `SynchronousNonBlockingStepExecution`?

jglick: What Remoting call? https://github.com/jenkinsci/jenkins/blob/8337315cced961853eb0d9f24034290066783662/core/src/main/java/hudson/Launcher.java#L1076-L1079

jglick:
> I have not really ever needed to write code like this

Actually we do this routinely in the Jenkins project: https://github.com/jenkins-infra/pipeline-library/blob/1a9f1d7251184abe4893f5d0231e206b2ee2ba66/vars/infra.groovy#L138-L142

But the point there is simply that there is not yet any single Pipeline durable task step which works on both Unix and Windows. `sh` works on Mac OS X (I hope!), so even assuming you had some stuff that you wanted to run on both Linux and Mac OS X, you would simply run a single `sh` command and then have that executable do any further discrimination below the Jenkins level, using standard tools like `uname`. So I see no need for a step like this.

jglick:
I would also suggest using, say, the `pipeline-utility-steps` plugin for assorted steps of occasional utility which core Pipeline devs would not need to commit to maintaining in perpetuity.

res0nance:
The use case was simply so that it is easier to hide implementation details and help code reuse for example. We can however do some trickery to determine the os via `uname` as @jglick suggested but this would be more convenient as this would just return a boolean. OTOH a pipeline-library-step probably would suffice.

```
def build() {
    // sequence of identical steps
   /**
       do mac specific thing
   */
    if (isMac()) {
        sh ''
    }
}
```

jglick:
KISS:

```groovy
def build() {
  sh 'bash build.sh'
}
```

```sh
# sequence of identical steps
# https://stackoverflow.com/a/18434831/12916 etc.
if [[ "$OSTYPE" == "darwin"* ]]; then some-mac-cmd; fi
```

dwnusbaum:
Given the discussion here, I am going to close this. Feel free to resubmit it against `https://github.com/jenkinsci/pipeline-utility-steps-plugin` or ask the maintainers of that plugin if they are interested in this step. If anyone finds a compelling reason for why this should be in workflow-basic-steps, feel free to reopen the PR.

PR #116, opened 2020-06-07
[JENKINS-29144] Extend CoreStep to also pass EnvVars to SimpleBuildStep.

Zastai:
See [JENKINS-29144](https://issues.jenkins-ci.org/browse/JENKINS-29144)

Builds on https://github.com/jenkinsci/jenkins/pull/4766 (incremental version 2.240-rc30032.5a9c198e2a33).

Note: I'm getting test failures from CatchErrorStepTest.optionalMessage(), but that doesn't involve CoreStep, so is probably caused by something else in current Jenkins master.

@jglick

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.1</version>
+        <version>4.2</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.21</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.241</jenkins.version>
+        <jenkins.version>2.164.3</jenkins.version>
         <java.level>8</java.level>
         <useBeta>true</useBeta>
     </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.56</version>
+        <version>4.1</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -80,26 +80,6 @@
         </dependencies>
     </dependencyManagement>
     <dependencies>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-jdk14</artifactId>
-            <version>1.7.26</version>
-        </dependency>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
-            <version>1.7.26</version>
-        </dependency>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>jcl-over-slf4j</artifactId>
-            <version>1.7.26</version>
-        </dependency>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>log4j-over-slf4j</artifactId>
-            <version>1.7.26</version>
-        </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.21</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.240-rc30032.5a9c198e2a33</jenkins.version>
+        <jenkins.version>2.241</jenkins.version>
         <java.level>8</java.level>
         <useBeta>true</useBeta>
     </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.21</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.164.3</jenkins.version>
+        <jenkins.version>2.240-rc30032.5a9c198e2a33</jenkins.version>
         <java.level>8</java.level>
         <useBeta>true</useBeta>
     </properties>
@@ -80,6 +80,26 @@
         </dependencies>
     </dependencyManagement>
     <dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-jdk14</artifactId>
+            <version>1.7.26</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.26</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+            <version>1.7.26</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>log4j-over-slf4j</artifactId>
+            <version>1.7.26</version>
+        </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/PwdStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/PwdStep.java
@@ -95,7 +95,11 @@ public class PwdStep extends Step {
 
         @Override protected String run() throws Exception {
             FilePath cwd = getContext().get(FilePath.class);
-            return (tmp ? WorkspaceList.tempDir(cwd) : cwd).getRemote();
+            if (tmp && cwd != null)
+                cwd = WorkspaceList.tempDir(cwd);
+            if (cwd == null)
+                return null; // FIXME: Or throw an exception?
+            return cwd.getRemote();
         }
 
         private static final long serialVersionUID = 1L;
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/PwdStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/PwdStep.java
@@ -28,8 +28,10 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.slaves.WorkspaceList;
+import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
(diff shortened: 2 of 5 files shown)


Zastai:
And a spotbugs error in code I have not touched. Yay. Should I fix these as part of this PR?

Zastai:
The remaining issue now, a test failure in CatchErrorStepTest.optionalMessage. That seems to test whether `catchError('foo')` works too (as opposed to mentioning the field name `catchError(message: 'foo')`).
But given that `CatchErrorStep` has multiple `String`-typed fields (the others being `buildResult` and `stageResult`), I'm not sure why that's expected to work.
As such, I don't know what to do to make that test work.

src/test/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStepTest.java
```diff
@@ -110,6 +111,7 @@
         r.assertLogContains("execution continued", b);
     }
 
+    @Ignore("This fails with recent Jenkins versions; not sure if it should still be expected to work.")
```
dwnusbaum, line 114: Hmm, I would've expected it to work, but I just checked and the test seems to have been passing for the wrong reason previously (the message "caught error" was part of an unexpected exception...), so it was actually already broken in the ways that matter. I am surprised I haven't seen any reports from users about this issue. This test being broken means that the step lost backwards compatibility in version 2.16 of this plugin for anyone using this syntax (we could probably fix it with a custom override of `StepDescriptor.newInstance`, but I'm not sure if it really matters, and that would be something for a different PR anyway). 

My best guess as to why the test is failing after the core update is something related to Pipeline logging and build completion, perhaps updating the parent POM would also fix the issue by pulling in test harness changes like https://github.com/jenkinsci/jenkins-test-harness/pull/210.

src/main/java/org/jenkinsci/plugins/workflow/steps/PwdStep.java
```diff
@@ -95,7 +95,11 @@ public boolean isTmp() {
 
         @Override protected String run() throws Exception {
             FilePath cwd = getContext().get(FilePath.class);
-            return (tmp ? WorkspaceList.tempDir(cwd) : cwd).getRemote();
+            if (tmp && cwd != null)
+                cwd = WorkspaceList.tempDir(cwd);
+            if (cwd == null)
```
dwnusbaum, line 100: This should never happen because the Pipeline engine uses `DescriptorImpl.getRequiredContext` (which returns a list containing `FilePath.class`) to check that all required context variables are available before the step is executed, and if they are not available, it throws an exception. I would either suppress the warning, or add `assert cwd != null` or `Objects.requireNonNull(cwd)` right after the `getContext()` call and leave the old line the same as it was.

pom.xml
```diff
@@ -80,6 +80,26 @@
...
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+            <version>1.7.26</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>log4j-over-slf4j</artifactId>
+            <version>1.7.26</version>
+        </dependency>
```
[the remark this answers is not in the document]
Zastai: OK, I'll give that a try once my working hours are over and see where I get.

src/test/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStepTest.java
Zastai: After updating to the 4.1 parent, I'll retry the test; if that makes it pass, I'll drop the `@Ignore`. Otherwise I'll leave it to prevent blocking builds of this PR, and a separate PR will be needed to either drop the test or fix the "pass the message as only positional param" feature.

src/main/java/org/jenkinsci/plugins/workflow/steps/PwdStep.java
Zastai: `WorkspaceList.tempDir()` is marked as `@CheckForNull`, so even with the `cwd` param checked for null, that may still trip the spotbugs check. Will give it a try.

dwnusbaum: Ah I forgot that `WorkspaceList.tempDir` is `@CheckForNull`. It should only return null if `cwd` does not have a parent directory. I think it would make sense to throw an exception if `tmp` is `true` and `cwd` has no parent directory since there is no temporary directory in that case (because it is supposed to be a sibling of the current directory).

pom.xml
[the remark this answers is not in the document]
Zastai: This did work. Unfortunately it also messes up IntelliJ, which can't resolve dependencies anymore (probably because it doesn't handle .mvn/maven.config). But command-line `mvn` still works.

src/test/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStepTest.java
Zastai: With the 4.1 parent this indeed passes on my end, probably because the
```
java.lang.IllegalArgumentException: Expected named arguments but got caught error
```
now ends up in the log again (and that matches the `caught error` string check). So I dropped the `@Ignore`.

src/main/java/org/jenkinsci/plugins/workflow/steps/PwdStep.java
Zastai: Added the `requireNonNull()`, and an `IOException` when `tempDir()` returns `null`. I did not add a Messages resources for that.

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -154,4 +168,36 @@
...
+            @Override
+            public boolean isApplicable(Class<? extends AbstractProject> jobType) {
+                return true;
+            }
+
+        }
+
+    }
+
+    @Test
```
jglick, line 196:
```suggestion
    @Issue("JENKINS-29144")
    @Test
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -154,4 +168,36 @@
...
 
+    public static class BuilderWithEnvironment extends Builder implements SimpleBuildStep {
+
+        @DataBoundConstructor
+        public BuilderWithEnvironment() {
+        }
+
+        @Override
+        public void perform(@Nonnull Run<?, ?> run, @Nonnull FilePath workspace, @Nonnull EnvVars env, @Nonnull Launcher launcher, @Nonnull TaskListener listener) throws InterruptedException, IOException {
+            assertNull(env.get("BUILD_ID"));
```
jglick, line 179: How is this `null` rather than `""`?

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -154,4 +168,36 @@
...
+
+        }
+
+    }
+
+    @Test
+    public void builderWithEnvironment() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node { withEnv(['TICKET=JENKINS-29144', 'BUILD_ID=']) { buildWithEnvironment() } }", true));
+        WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
```
jglick, line 200:
```suggestion
        WorkflowRun b = r.buildAndAssertSuccess(p);
```

or even

```suggestion
        r.buildAndAssertSuccess(p);
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
Zastai: Well, the docs for withEnv say that "FOO=" removes it from the environment, so I assumed it would be null. The test seems to agree :)

jglick: Huh, https://github.com/jenkinsci/pipeline-plugin/commit/5b7d046506f0f52480328e218e906583792b4bae five years ago :man_shrugging:

Zastai:
So what happens next? Wait for jenkins/PR4766 to be merged, then when the weekly is released, reference that and convert this to non-draft?

dwnusbaum:
> Wait for jenkins/PR4766 to be merged, then when the weekly is released, reference that and convert this to non-draft?

Yes, more or less. One thing though is that this plugin is used widely enough that I do not want it to depend on such a new version of Jenkins core, instead I would prefer to use reflection to call the new overload of `perform` only when it is available and otherwise call the existing overload. You can drop the `@Override` annotation in the test so that things still compile against an older version of Jenkins core and I think you can use `Jenkins.getVersion` to change the assertions depending on what version of Jenkins is being tested against, and I can push a commit to your branch to update the Jenkinsfile to build the plugin against both the current baseline and the weekly release with your change to test the new behavior.

Zastai:
That's a bit messy, and it'll get messier when adding [JENKINS-46175](https://issues.jenkins-ci.org/browse/JENKINS-46175).

I thought to have read something about plugin versioning supporting "branches", so 2.22 and onwards would reference jenkins 2.240 (or whatever the weekly turns out to be), but a later 2.25.1 might reference 2.163 (for a fix for a bug reported for 2.21 or earlier that does not depend on new API), with the update center offering the 2.21 -> 2.25.1 upgrade to users running older Jenkins releases.
But I may have misunderstood.

jglick:
> I thought to have read something about plugin versioning supporting "branches"

Yes, my preference (FWIW) is to keep things simple, declaring whatever Jenkins version you want, and using something like https://gist.github.com/jglick/86a30894446ed38f918050c1180483e2 to create backport branches on demand (typically just for significant bug fixes). But the decision rests on whomever is actually merging PRs and cutting releases, probably @dwnusbaum in this case.

Zastai:
Now that the core part of JENKINS-21944 is merged, I can start JENKINS-46175.
Given they affect the same code area here, I wonder whether it would make sense to use this PR to apply both sets of changes here, especially if I'd have to go around via reflection to access the new APIs.

Zastai:
Decided to keep this separate despite the overlap with #117

Zastai:
Switched to the released 2.241.

Then did as requested and dropped the requirement back to 2.164.3, invoking the new API via reflection when running on 2.241 or later.

Would just need that Jenkinsfile change to build against old & new.

Zastai:
The build also seems to have failed for no obvious reason.

Zastai:
Closing - superseded by https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/121

jglick:
Note that you do not need to close this: if #121 is merged (without squash or rebase!), this will automatically be marked merged as well.

Zastai:
I did not know that. No harm done I guess.

PR #118, opened 2020-06-18
Add new step 'debug'

fcappi:
Hi,

I'm submitting this PR to create a new simple common step for debug purpose. It will basically only execute the Closure block if the parameter is TRUE.

```
debug(true) {
 < all this block will be executed >
}

debug(false) {
 < all this block will be skipped >
}
```

I think this will improve the readability of the pipelines, cause the developer can use it to make it clear that all that block is intended to output debug data and can be ignored by others just trying to understand the logic executed by the pipeline.

The parameter allows the developer to define when debug mode is enabled or not(for instance, it could be a pipeline parameter).

Signed-off-by: Fernando Cappi <3818944+fcappi@users.noreply.github.com>

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/DebugStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/DebugStep.java
@@ -1,0 +1,88 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2014, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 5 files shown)


dwnusbaum:
@fcappi Thanks for the PR! I would probably recommend something like this instead, which doesn't require any special steps:

```
def DEBUG = ...
if (DEBUG) {
  ...
}
```

Is there any particular reason you want to create a step for this use case instead of using `if`?

fcappi:
Hi @dwnusbaum , thanks for the comment!

I consider the following as benefits of using a step instead of IF statement:
- I think it makes the developer intent more clear in scenarios where the condition is not so straightforward:
Example
```
debug(<exeptional scenario occured>) {
<do extra steps>
}
```
- If used as a standard way of adding debug code to the pipeline:
  - Allows us to easily find all debug related code with a simple search;
  - Allow user to overwrite the step if needed, applying a common logic (e.g. standard log output) to all debug blocks;
- All debug related logs are consolidate per block and can be reviewed separated in the Jenkins UI;

Anyways, we are not forcing the developers to use it so it would be just an alternative way of doing that, with the differences as listed above.

Please let me know your thoughts about that!

fcappi:
Also for scripted pipelines, it makes the code less verbose:

```
script {
  if(DEBUG) {
    //do 
  }
}
```
vs
```
debug(DEBUG) {
  //do 
}
```

dwnusbaum:
Thanks for adding some details about your use case. I don't think it's something we will add to this plugin. You could create a new plugin for it, or maybe see if the maintainers of https://github.com/jenkinsci/pipeline-utility-steps-plugin are interested in it.

Note that in general I am not inclined to add any new steps to this plugin except for special cases, (for example if the step is closely related to a step in this plugin already, or replaces a popular feature that is not a step like `currentBuild`) so that core Pipeline maintainers are not responsible for maintaining new steps forever (for example we also rejected #109 for similar reasons). I should probably update the PR template or README or something to be more explicit about this.

> Also for scripted pipelines

I think you mean Declarative Pipelines, but FWIW I think that we probably would specifically not want to add a general conditional step like this to use in Declarative, because it kind of leads us down the path of creating steps for other control flow structures.

fcappi:
OK, thanks for the explanation and that makes sense.

I will close the PR.

Note: fixed scripted -> declarative in my comment above.

PR #119, opened 2020-06-19
Update Helper text with max wait time

lionelraeymaekers:
Add clarification to the maximum wait time between retries in the helper

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep/help.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep/help.html
@@ -1,7 +1,7 @@
 <div>
     Runs its body repeatedly until it returns <code>true</code>.
     If it returns <code>false</code>, waits a while and tries again.
-    (Subsequent failures will slow down the delay between attempts.)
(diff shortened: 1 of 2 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep/help.html
```diff
@@ -4,4 +4,5 @@
     (Subsequent failures will slow down the delay between attempts.)
```
dwnusbaum, line 4:
```suggestion
    (Subsequent failures will slow down the delay between attempts, up to a maximum of 15 seconds.)
```

src/main/resources/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep/help.html
```diff
@@ -4,4 +4,5 @@
...
     There is no limit to the number of retries,
     but if the body throws an error that is thrown up immediately.
+    The maximum wait time between attempt is 15 seconds (even if you set a recurrence period higher)
```
dwnusbaum, line 7:
```suggestion
```

PR #121, opened 2020-07-02
[JENKINS-29144] Extend CoreStep to also pass EnvVars to SimpleBuildStep

dwnusbaum:
Subsumes #116. See [JENKINS-29144](https://issues.jenkins-ci.org/browse/JENKINS-29144).

This PR adds some minor formatting changes, updates the Jenkinsfile to test against 2.241 as well as the current baseline, and changes `CoreStep` from using the Jenkins version to decide whether to call the new overload of `perform` to directly check for the new method instead.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
@@ -37,7 +37,6 @@ import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.tasks.Builder;
 import hudson.tasks.Publisher;
-import hudson.util.VersionNumber;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
```diff
@@ -75,9 +78,21 @@
...
-            delegate.perform(getContext().get(Run.class), workspace, getContext().get(Launcher.class), getContext().get(TaskListener.class));
+            final Run<?,?> run = Objects.requireNonNull(ctx.get(Run.class));
+            final Launcher launcher = Objects.requireNonNull(ctx.get(Launcher.class));
+            final TaskListener listener = Objects.requireNonNull(ctx.get(TaskListener.class));
+            // TODO: Replace with a direct call to SimpleBuildStep.perform(Run, FilePath, EnvVars, Launcher, TaskListener) once the minimum core version for this plugin is 2.241 or newer.
+            try {
+                final Method perform = this.delegate.getClass().getMethod("perform", Run.class, FilePath.class,
+                        EnvVars.class, Launcher.class, TaskListener.class);
+                final EnvVars env = Objects.requireNonNull(ctx.get(EnvVars.class));
+                perform.invoke(this.delegate, run, workspace, env, launcher, listener);
```
dwnusbaum, line 92: I prefer to check for the method directly instead of having an indirect check against the Jenkins version. Notably though, this means that the new API would be active in Pipelines even when running old versions of Jenkins core if a `SimpleBuildStep` implements the new method (as seen in the newly-added test case, which now runs against old versions of Jenkins), which could create some confusion.

Zastai: Seems fine to me. If we're running on 2.241 or later, that'll be expected behaviour. And if running on older Jenkins, that other plugin would have had to implement the new API without pulling in the corresponding declarations, which feels like a bit of an edge case.

Zastai:
Might make sense to use JDK8 for the 2.241 build branch too for now; the failure with JDK11 is separate from the changes here, so should probably handled in a separate PR.

Zastai:
What's more, I think that spotbugs error is bogus - it's complaining about code generated by the compiler for try-with-resources.
It essentially emits "`if (os != null) os.close();`" (leaving it to the JIT to elide the null test), and spotbugs says "but if it was null the `os.write()` would have thrown a NPE".

dwnusbaum:
> Might make sense to use JDK8 for the 2.241 build branch too for now

Yeah good point, I'll look into the issue and decide what to do.

> What's more, I think that spotbugs error is bogus ...

Yeah, that is https://github.com/spotbugs/spotbugs/issues/756.

PR #122, opened 2020-07-13
PwdStep: Handle possible null return from WorkspaceList.tempDir().

Zastai:
`WorkspaceList.tempDir()` is marked `@CheckForNull` but the result is used unconditionally.

I figure it's better to get this merged on its own and then merge the outstanding PRs so they don't have unrelated changes.

PR #117, opened 2020-06-14
[JENKINS-46175] Make workspace and launcher optional for CoreStep and CoreWrapperStep.

Zastai:
This also includes the same `PwdStep` spotbugs fix as https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/116 does.

PR #123, opened 2020-07-13
[JENKINS-46175] Make workspace context optional for CoreStep and CoreWrapperStep.

Zastai:
This is a variation of https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/117, using the API changes from https://github.com/jenkinsci/jenkins/pull/4829.

Also contains the `PwdStep` fix from https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/122.

Marking as draft because there are multiple variations, both here and at the core level.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
@@ -43,7 +43,6 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import javax.annotation.CheckForNull;
 import jenkins.model.Jenkins;
@@ -88,7 +87,11 @@ public final class CoreStep extends Step {
             final FilePath workspace = ctx.get(FilePath.class);
             final Launcher launcher = ctx.get(Launcher.class);
             boolean workspaceRequired = true;
-            // TODO: Replace with a direct call to SimpleBuildStep.requiresWorkspace() once the minimum core version for this plugin is 2.25x or newer.
+            // In Jenkins 2.25x, a SimpleBuildStep can indicate that it does not require a workspace context by
+            // overriding a requiresWorkspace() method to return false. So use that if it's available.
+            // Note: this uses getMethod() on the delegate's type and not SimpleBuildStep so that an implementation can
+            // get this behaviour without switching to Jenkins 2.25x itself.
+            // TODO: Use 'workspaceRequired = this.delegate.requiresWorkspace()' once this plugin depends on Jenkins 2.25x or later.
             try {
                 final Method requiresWorkspace = this.delegate.getClass().getMethod("requiresWorkspace");
                 workspaceRequired = (boolean) requiresWorkspace.invoke(this.delegate);
@@ -108,7 +111,11 @@ public final class CoreStep extends Step {
             }
             // always pass the workspace context when available, even when it is not strictly required
             if (workspace != null && launcher != null) {
-                // TODO: Replace with a direct call to SimpleBuildStep.perform(Run, FilePath, EnvVars, Launcher, TaskListener) once the minimum core version for this plugin is 2.241 or newer.
+                // In Jenkins 2.241, a SimpleBuildStep has an overload of perform() that also takes an EnvVars. If that
+                // is available, we use it.
+                // Note: this uses getMethod() on the delegate's type and not SimpleBuildStep so that an implementation
+                // can get this behaviour without switching to Jenkins 2.241 itself.
+                // TODO: Use 'this.delegate.perform(run, workspace, env, launcher, listener)' once this plugin depends on Jenkins 2.241 or later.
                 try {
                     final Method perform = this.delegate.getClass().getMethod("perform", Run.class, FilePath.class,
                             EnvVars.class, Launcher.class, TaskListener.class);
@@ -117,8 +124,11 @@ public final class CoreStep extends Step {
                     this.delegate.perform(run, workspace, launcher, listener);
                 }
             } else {
-                // TODO: Replace with a direct call to SimpleBuildStep.perform(Run, EnvVars, TaskListener) once the minimum core version for this plugin is 2.25x or newer.
-                // Note: no try here; if we get here, the method MUST exist
+                // If we get here, workspaceRequired is false and there is no workspace context. In that case, the
+                // overload of perform() introduced in Jenkins 2.25x MUST exist (so no try block here).
+                // Note: this uses getMethod() on the delegate's type and not SimpleBuildStep so that an implementation
+                // can get this behaviour without switching to Jenkins 2.25x itself.
+                // TODO: Use 'this.delegate.perform(run, env, listener)' once the minimum core version for this plugin is 2.25x or newer.
                 final Method perform = this.delegate.getClass().getMethod("perform", Run.class, EnvVars.class, TaskListener.class);
                 perform.invoke(this.delegate, run, env, listener);
             }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
@@ -27,6 +27,7 @@ package org.jenkinsci.plugins.workflow.steps;
 import com.google.common.collect.ImmutableSet;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
+import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
@@ -75,9 +76,32 @@ public final class CoreStep extends Step {
         }
 
         @Override protected Void run() throws Exception {
-            FilePath workspace = getContext().get(FilePath.class);
-            workspace.mkdirs();
-            delegate.perform(getContext().get(Run.class), workspace, getContext().get(Launcher.class), getContext().get(TaskListener.class));
+            final StepContext context = getContext();
+            final Run<?,?> run = context.get(Run.class);
+            assert run != null;
+            final TaskListener listener = context.get(TaskListener.class);
+            assert listener != null;
+            final EnvVars env = context.get(EnvVars.class);
+            assert env != null;
+            final FilePath workspace = context.get(FilePath.class);
+            final Launcher launcher = context.get(Launcher.class);
+            if (this.delegate.requiresWorkspace()) {
+                if (workspace == null) {
+                    throw new MissingContextVariableException(FilePath.class);
+                }
+                if (launcher == null) {
+                    throw new MissingContextVariableException(Launcher.class);
+                }
+            }
+            if (workspace != null) {
+                workspace.mkdirs();
+            }
+            // always pass the workspace context when available, even when it is not strictly required
+            if (workspace != null && launcher != null) {
+                this.delegate.perform(run, workspace, env, launcher, listener);
+            } else {
+                this.delegate.perform(run, env, listener);
+            }
             return null;
         }
 
@@ -121,7 +145,7 @@ public final class CoreStep extends Step {
         }
 
         @Override public Set<? extends Class<?>> getRequiredContext() {
-            return ImmutableSet.of(Run.class, FilePath.class, Launcher.class, TaskListener.class);
+            return ImmutableSet.of(Run.class, TaskListener.class);
         }
 
         @Override public String argumentsToString(Map<String, Object> namedArgs) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
@@ -37,6 +37,7 @@ import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.tasks.Builder;
 import hudson.tasks.Publisher;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -97,6 +98,9 @@ public final class CoreStep extends Step {
                 workspaceRequired = (boolean) requiresWorkspace.invoke(this.delegate);
             } catch(NoSuchMethodException e) {
                 // ok, default to true
+            } catch (InvocationTargetException ite) {
+                final Throwable realException = ite.getCause();
+                throw realException instanceof Exception ? (Exception) realException : ite;
             }
             if (workspaceRequired) {
                 if (workspace == null) {
@@ -122,15 +126,23 @@ public final class CoreStep extends Step {
                     perform.invoke(this.delegate, run, workspace, env, launcher, listener);
                 } catch (NoSuchMethodException e) {
                     this.delegate.perform(run, workspace, launcher, listener);
+                } catch (InvocationTargetException ite) {
+                    final Throwable realException = ite.getCause();
+                    throw realException instanceof Exception ? (Exception) realException : ite;
                 }
             } else {
                 // If we get here, workspaceRequired is false and there is no workspace context. In that case, the
-                // overload of perform() introduced in Jenkins 2.25x MUST exist (so no try block here).
+                // overload of perform() introduced in Jenkins 2.25x MUST exist.
                 // Note: this uses getMethod() on the delegate's type and not SimpleBuildStep so that an implementation
                 // can get this behaviour without switching to Jenkins 2.25x itself.
                 // TODO: Use 'this.delegate.perform(run, env, listener)' once the minimum core version for this plugin is 2.25x or newer.
                 final Method perform = this.delegate.getClass().getMethod("perform", Run.class, EnvVars.class, TaskListener.class);
-                perform.invoke(this.delegate, run, env, listener);
+                try {
+                    perform.invoke(this.delegate, run, env, listener);
+                } catch (InvocationTargetException ite) {
+                    final Throwable realException = ite.getCause();
+                    throw realException instanceof Exception ? (Exception) realException : ite;
+                }
             }
             return null;
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
@@ -86,11 +86,11 @@ public final class CoreStep extends Step {
             final TaskListener listener = Objects.requireNonNull(ctx.get(TaskListener.class));
             final EnvVars env = Objects.requireNonNull(ctx.get(EnvVars.class));
             boolean workspaceRequired = true;
-            // In Jenkins 2.25x, a SimpleBuildStep can indicate that it does not require a workspace context by
+            // In Jenkins 2.258, a SimpleBuildStep can indicate that it does not require a workspace context by
             // overriding a requiresWorkspace() method to return false. So use that if it's available.
             // Note: this uses getMethod() on the delegate's type and not SimpleBuildStep so that an implementation can
-            // get this behaviour without switching to Jenkins 2.25x itself.
-            // TODO: Use 'workspaceRequired = this.delegate.requiresWorkspace()' once this plugin depends on Jenkins 2.25x or later.
+            // get this behaviour without switching to Jenkins 2.258 itself.
+            // TODO: Use 'workspaceRequired = this.delegate.requiresWorkspace()' once this plugin depends on Jenkins 2.258 or later.
             try {
                 final Method requiresWorkspace = this.delegate.getClass().getMethod("requiresWorkspace");
                 workspaceRequired = (boolean) requiresWorkspace.invoke(this.delegate);
@@ -116,10 +116,10 @@ public final class CoreStep extends Step {
                 delegate.perform(run, workspace, env, launcher, listener);
             } else {
                 // If we get here, workspaceRequired is false and there is no workspace context. In that case, the
-                // overload of perform() introduced in Jenkins 2.25x MUST exist.
+                // overload of perform() introduced in Jenkins 2.258 MUST exist.
                 // Note: this uses getMethod() on the delegate's type and not SimpleBuildStep so that an implementation
-                // can get this behaviour without switching to Jenkins 2.25x itself.
-                // TODO: Use 'this.delegate.perform(run, env, listener)' once the minimum core version for this plugin is 2.25x or newer.
+                // can get this behaviour without switching to Jenkins 2.258 itself.
+                // TODO: Use 'this.delegate.perform(run, env, listener)' once the minimum core version for this plugin is 2.258 or newer.
                 final Method perform = this.delegate.getClass().getMethod("perform", Run.class, EnvVars.class, TaskListener.class);
                 try {
                     perform.invoke(this.delegate, run, env, listener);
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
@@ -93,19 +93,43 @@ public class CoreWrapperStep extends Step {
 
         private void doStart() throws Exception {
             SimpleBuildWrapper.Context c = new SimpleBuildWrapper.Context();
-            Run<?, ?> run = getContext().get(Run.class);
-            delegate.setUp(c, run, getContext().get(FilePath.class), getContext().get(Launcher.class), getContext().get(TaskListener.class), getContext().get(EnvVars.class));
-            BodyInvoker bodyInvoker = getContext().newBodyInvoker();
+            c.setWorkspaceRequirement(this.delegate);
+            final StepContext context = getContext();
+            final Run<?, ?> run = context.get(Run.class);
+            assert run != null;
+            {
+                final TaskListener listener = context.get(TaskListener.class);
+                assert listener != null;
+                final EnvVars env = context.get(EnvVars.class);
+                assert env != null;
+                final FilePath workspace = context.get(FilePath.class);
+                final Launcher launcher = context.get(Launcher.class);
+                if (this.delegate.requiresWorkspace()) {
+                    if (workspace == null) {
+                        throw new MissingContextVariableException(FilePath.class);
+                    }
+                    if (launcher == null) {
+                        throw new MissingContextVariableException(Launcher.class);
+                    }
+                }
+                // always pass the workspace context when available, even when it is not strictly required
+                if (workspace != null && launcher != null) {
+                    this.delegate.setUp(c, run, workspace, launcher, listener, env);
+                } else {
+                    this.delegate.setUp(c, run, listener, env);
+                }
+            }
+            BodyInvoker bodyInvoker = context.newBodyInvoker();
             Map<String,String> overrides = c.getEnv();
             if (!overrides.isEmpty()) {
-                bodyInvoker.withContext(EnvironmentExpander.merge(getContext().get(EnvironmentExpander.class), new ExpanderImpl(overrides)));
+                bodyInvoker.withContext(EnvironmentExpander.merge(context.get(EnvironmentExpander.class), new ExpanderImpl(overrides)));
             }
             ConsoleLogFilter filter = delegate.createLoggerDecorator(run);
             if (filter != null) {
-                bodyInvoker.withContext(BodyInvoker.mergeConsoleLogFilters(getContext().get(ConsoleLogFilter.class), filter));
+                bodyInvoker.withContext(BodyInvoker.mergeConsoleLogFilters(context.get(ConsoleLogFilter.class), filter));
             }
             SimpleBuildWrapper.Disposer disposer = c.getDisposer();
-            bodyInvoker.withCallback(disposer != null ? new Callback2(disposer) : BodyExecutionCallback.wrap(getContext())).start();
+            bodyInvoker.withCallback(disposer != null ? new Callback2(disposer) : BodyExecutionCallback.wrap(context)).start();
         }
 
         private final class Callback2 extends TailCall {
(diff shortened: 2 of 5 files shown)


Zastai:
(Sorry about the force-pushes; I messed up a few merges.)

jglick:
Some sort of CI issues…

dwnusbaum:
> Some sort of CI issues…

Yeah IDK what the root cause is, but I filed https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/124 to investigate.

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
```diff
@@ -78,16 +78,24 @@
...
             assert env != null;
-            final FilePath workspace = context.get(FilePath.class);
-            final Launcher launcher = context.get(Launcher.class);
-            if (this.delegate.requiresWorkspace()) {
+            final FilePath workspace = ctx.get(FilePath.class);
+            final Launcher launcher = ctx.get(Launcher.class);
+            boolean workspaceRequired = true;
+            // TODO: Replace with a direct call to SimpleBuildStep.requiresWorkspace() once the minimum core version for this plugin is 2.25x or newer.
+            try {
+                final Method requiresWorkspace = this.delegate.getClass().getMethod("requiresWorkspace");
```
jglick, line 93:
```suggestion
                final Method requiresWorkspace = SimpleBuildStep.class.getMethod("requiresWorkspace");
```

and similarly elsewhere

Zastai: Wouldn’t that always call the default implementation instead of an override in the actual step instance?
I could see this for the `Disposer` case because that has a non-virtual one, but not for the `requiresWorkspace` on step and wrapper.

jglick: The `Method` is defined in a particular class: `SimpleBuildStep`. `invoke` will run the implementation. There is no way in Java to do otherwise, other than by generating bytecode that calls `invokespecial`.

Zastai: No, `SimpleBuildStep` is an interface with a default implementation for `requiresWorkspace`. If you use `getMethod` on `SimpleBuildStep.class`, that default implementation is what you get and invoke, This breaks the processing (and the without-workspace-requirement test), because on an older Jenkins it will always throw `NoSuchMethodException`, causing it to require a workspace, and on a new Jenkins it always returns `true`, with the same effect.

It is essential to call the `getMethod("requiresWorkspace")` on the actual step type captured in `this.delegate`, so that it will find the override and invoke that.

The only case where this is not as important is on the Disposer(), because in new Jenkins, where the method exists, it is final. However, that causes a problem for the reflection-based approach, because that prevents any unit test for a case of a wrapper+disposer not requiring a workspace, because the disposer invocation would always think it required a workspace (due to the method not existing in `Disposer` itself). Given that I don't think a subclass can define a new method with the same signature as a final method from a superclass, also using the specific disposer type for the lookup should be harmless, and allows the unit test to work (and allows implementers of wrappers not to require a workspace when used in pipelines, even on old Jenkins).

Zastai:
Un-drafting this. The way it's set up now, it does not actually depend on https://github.com/jenkinsci/jenkins/pull/4829 - but it will make use of the API if it's available.

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
jglick: Really? Because

```java
import java.lang.reflect.Method;
public class M {
    public static void main(String[] args) throws Exception {
        Method m = I.class.getMethod("m");
        m.invoke(new C1());
        m.invoke(new C2());
    }
    interface I {
        default void m() {
            System.out.println("from I");
        }
    }
    static class C1 implements I {}
    static class C2 implements I {
        @Override public void m() {
            System.out.println("from C2");
        }
    }
}
```

prints

```
from I
from C2
```

as I would expect. Am I missing something?

Zastai: Hmmm - it may just be the unit tests that get broken then (because they have "fake" overrides).

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -204,4 +204,88 @@ public void builderWithEnvironment() throws Exception {
...
+            @Override
+            public boolean isApplicable(Class<? extends AbstractProject> jobType) {
+                return true;
+            }
+        }
+    }
+
+    @Issue("JENKINS-46175")
+    @Test
+    public void builderWithWorkspaceRequirement() throws Exception {
```
jglick, line 237: All these new tests should, say,

```java
try {
    SimpleBuildStep.class.getMethod("requiresWorkspace");
} catch (NoSuchMethodException x) {
    assumeNoException("skipping on old Jenkins", x);
}
```

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
jglick: Yes I think that was the disconnect: https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/123#discussion_r461755336

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
Zastai: Well, for the environment PR, ISTR one of the goals stated by @dwnusbaum was to allow implementers to do like these tests do, and essentially implement the new API without needing the new Jenkins.
So foo-plugin could set things up not to require a workspace, enabling use outside of a node context with a recent workflow-basic-steps, even on old Jenkins.
That's supported perfectly well by the implementation as it stands (as evidenced by the unit tests, which work fine on old Jenkins).

jglick: > one of the goals stated by @dwnusbaum…

Oh? I must have missed that. We occasionally do this, though it is a bit messy in this case because you are still obliged to implement the old overloads.

Zastai: I'd have to go back and see if I remember it correctly.
I mean, I don't mind removing this as a feature, but it does not seem to cost much on this end, so I figure it makes sense to keep it. Unless the current approach actually breaks in any scenarios with New Jenkins - but I don't think it does.

jglick: OK, no strong opinion either way, but I would recommend adding comments to the points where you call `xxx.getClass().getMethod(…)` to explain that

* in old Jenkins, this will either return a method defined in some concrete class, or throw NSME
* in new Jenkins, this will return the interface method (or an override, but the behavior is the same)

Zastai: I've added more comments, but only described it in general terms. Hope that is enough.

Zastai:
I'm getting one test failure locally on CoreStepTest.mailer (`ClassNotFoundException: org.jenkinsci.plugins.displayurlapi.DisplayURLProvider`); this seems to be because the mailer plugin now has a new dependency, so this PR didn't seem like the right place to fix that.

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
```diff
@@ -78,13 +80,54 @@
...
+            final Launcher launcher = ctx.get(Launcher.class);
             final TaskListener listener = Objects.requireNonNull(ctx.get(TaskListener.class));
             final EnvVars env = Objects.requireNonNull(ctx.get(EnvVars.class));
-            delegate.perform(run, workspace, env, launcher, listener);
+            boolean workspaceRequired = true;
+            // In Jenkins 2.25x, a SimpleBuildStep can indicate that it does not require a workspace context by
+            // overriding a requiresWorkspace() method to return false. So use that if it's available.
+            // Note: this uses getMethod() on the delegate's type and not SimpleBuildStep so that an implementation can
+            // get this behaviour without switching to Jenkins 2.25x itself.
+            // TODO: Use 'workspaceRequired = this.delegate.requiresWorkspace()' once this plugin depends on Jenkins 2.25x or later.
```
jglick, line 93:
```suggestion
            // get this behaviour without switching to Jenkins 2.258 itself.
            // TODO: Use 'workspaceRequired = this.delegate.requiresWorkspace()' once this plugin depends on Jenkins 2.258 or later.
```

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreStep.java
```diff
@@ -78,13 +80,54 @@
...
+            }
+            // always pass the workspace context when available, even when it is not strictly required
+            if (workspace != null && launcher != null) {
+                delegate.perform(run, workspace, env, launcher, listener);
+            } else {
+                // If we get here, workspaceRequired is false and there is no workspace context. In that case, the
+                // overload of perform() introduced in Jenkins 2.25x MUST exist.
+                // Note: this uses getMethod() on the delegate's type and not SimpleBuildStep so that an implementation
+                // can get this behaviour without switching to Jenkins 2.25x itself.
+                // TODO: Use 'this.delegate.perform(run, env, listener)' once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 122:
```suggestion
                // overload of perform() introduced in Jenkins 2.258 MUST exist.
                // Note: this uses getMethod() on the delegate's type and not SimpleBuildStep so that an implementation
                // can get this behaviour without switching to Jenkins 2.258 itself.
                // TODO: Use 'this.delegate.perform(run, env, listener)' once the minimum core version for this plugin is 2.258 or newer.
```

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
```diff
@@ -92,20 +94,82 @@ public SimpleBuildWrapper getDelegate() {
...
 
         private void doStart() throws Exception {
-            SimpleBuildWrapper.Context c = new SimpleBuildWrapper.Context();
-            Run<?, ?> run = getContext().get(Run.class);
-            delegate.setUp(c, run, getContext().get(FilePath.class), getContext().get(Launcher.class), getContext().get(TaskListener.class), getContext().get(EnvVars.class));
-            BodyInvoker bodyInvoker = getContext().newBodyInvoker();
+            SimpleBuildWrapper.Context c = null;
+            // In Jenkins 2.25x, a createContext() method on SimpleBuildWrapper is required to ensure that a Disposer
+            // registered on that context inherits the wrapper's workspace requirement. Use it when available.
+            // TODO: Use 'c = this.delegate.createContext()' once this plugin depends on Jenkins 2.25x or later.
```
jglick, line 100:
```suggestion
            // In Jenkins 2.258, a createContext() method on SimpleBuildWrapper is required to ensure that a Disposer
            // registered on that context inherits the wrapper's workspace requirement. Use it when available.
            // TODO: Use 'c = this.delegate.createContext()' once this plugin depends on Jenkins 2.258 or later.
```

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
```diff
@@ -92,20 +94,82 @@ public SimpleBuildWrapper getDelegate() {
...
+                final EnvVars env = context.get(EnvVars.class);
+                assert env != null;
+                final FilePath workspace = context.get(FilePath.class);
+                final Launcher launcher = context.get(Launcher.class);
+                boolean workspaceRequired = true;
+                // In Jenkins 2.25x, a SimpleBuildWrapper can indicate that it does not require a workspace context by
+                // overriding a requiresWorkspace() method to return false. So use that if it's available.
+                // Note: this uses getMethod() on the delegate's type and not SimpleBuildWrapper so that an
+                // implementation can get this behaviour without switching to Jenkins 2.25x itself.
+                // TODO: Use 'workspaceRequired = this.delegate.requiresWorkspace()' once this plugin depends on Jenkins 2.25x or later.
```
jglick, line 126:
```suggestion
                // In Jenkins 2.258, a SimpleBuildWrapper can indicate that it does not require a workspace context by
                // overriding a requiresWorkspace() method to return false. So use that if it's available.
                // Note: this uses getMethod() on the delegate's type and not SimpleBuildWrapper so that an
                // implementation can get this behaviour without switching to Jenkins 2.258 itself.
                // TODO: Use 'workspaceRequired = this.delegate.requiresWorkspace()' once this plugin depends on Jenkins 2.258 or later.
```

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
```diff
@@ -92,20 +94,82 @@ public SimpleBuildWrapper getDelegate() {
...
+                }
+                // always pass the workspace context when available, even when it is not strictly required
+                if (workspace != null && launcher != null) {
+                    this.delegate.setUp(c, run, workspace, launcher, listener, env);
+                } else {
+                    // If we get here, workspaceRequired is false and there is no workspace context. In that case, the
+                    // overload of setUp() introduced in Jenkins 2.25x MUST exist.
+                    // Note: this uses getMethod() on the delegate's type and not SimpleBuildWrapper so that an
+                    // implementation can get this behaviour without switching to Jenkins 2.25x itself.
+                    // TODO: Use 'this.delegate.setUp(c, run, listener, env)' once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 152:
```suggestion
                    // overload of setUp() introduced in Jenkins 2.258 MUST exist.
                    // Note: this uses getMethod() on the delegate's type and not SimpleBuildWrapper so that an
                    // implementation can get this behaviour without switching to Jenkins 2.258 itself.
                    // TODO: Use 'this.delegate.setUp(c, run, listener, env)' once the minimum core version for this plugin is 2.258 or newer.
```

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
```diff
@@ -149,7 +213,53 @@ private void doStart() throws Exception {
...
+            assert listener != null;
+            final FilePath workspace = context.get(FilePath.class);
+            final Launcher launcher = context.get(Launcher.class);
+            boolean workspaceRequired = true;
+            // In Jenkins 2.25x, a Disposer has a final requiresWorkspace() method that indicates whether or not it (or
+            // more accurately, its associated wrapper) requires a workspace context. This is set up via the Context, as
+            // long as that is created via SimpleBuildWrapper.createContext().
+            // Note: this uses getMethod() on the disposer's type and not SimpleBuildWrapper.Disposer so that an
+            // implementation can get this behaviour without switching to Jenkins 2.25x itself.
+            // TODO: Use 'workspaceRequired = this.disposer.requiresWorkspace()' once this plugin depends on Jenkins 2.25x or later.
```
jglick, line 228:
```suggestion
            // In Jenkins 2.258, a Disposer has a final requiresWorkspace() method that indicates whether or not it (or
            // more accurately, its associated wrapper) requires a workspace context. This is set up via the Context, as
            // long as that is created via SimpleBuildWrapper.createContext().
            // Note: this uses getMethod() on the disposer's type and not SimpleBuildWrapper.Disposer so that an
            // implementation can get this behaviour without switching to Jenkins 2.258 itself.
            // TODO: Use 'workspaceRequired = this.disposer.requiresWorkspace()' once this plugin depends on Jenkins 2.258 or later.
```

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
```diff
@@ -149,7 +213,53 @@ private void doStart() throws Exception {
...
+            }
+            // always pass the workspace context when available, even when it is not strictly required
+            if (workspace != null && launcher != null) {
+                this.disposer.tearDown(run, workspace, launcher, listener);
+            } else {
+                // If we get here, workspaceRequired is false and there is no workspace context. In that case, the
+                // overload of tearDown() introduced in Jenkins 2.25x MUST exist.
+                // Note: this uses getMethod() on the disposer's type and not SimpleBuildWrapper.Disposer so that an
+                // implementation can get this behaviour without switching to Jenkins 2.25x itself.
+                // TODO: Use 'this.disposer.tearDown(run, listener)' once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 254:
```suggestion
                // overload of tearDown() introduced in Jenkins 2.258 MUST exist.
                // Note: this uses getMethod() on the disposer's type and not SimpleBuildWrapper.Disposer so that an
                // implementation can get this behaviour without switching to Jenkins 2.258 itself.
                // TODO: Use 'this.disposer.tearDown(run, listener)' once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -203,4 +203,88 @@ public void builderWithEnvironment() throws Exception {
         r.buildAndAssertSuccess(p);
     }
 
+    public static class BuilderWithWorkspaceRequirement extends Builder implements SimpleBuildStep {
+        @DataBoundConstructor
+        public BuilderWithWorkspaceRequirement() {
+        }
+        // TODO: Remove once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 210:
```suggestion
        // TODO: Remove once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -203,4 +203,88 @@ public void builderWithEnvironment() throws Exception {
...
+        public void perform(@Nonnull Run<?, ?> run, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull TaskListener listener) throws InterruptedException, IOException {
+            fail("This method should not get called.");
+        }
+        // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 214:
```suggestion
        // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -203,4 +203,88 @@ public void builderWithEnvironment() throws Exception {
...
+        public void perform(@Nonnull Run<?, ?> run, @Nonnull EnvVars env, @Nonnull TaskListener listener) throws InterruptedException, IOException {
+            listener.getLogger().println("workspace context required, but not provided!");
+        }
+        // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 218:
```suggestion
        // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -203,4 +203,88 @@ public void builderWithEnvironment() throws Exception {
...
+        // and fails outside of one
+        p.setDefinition(new CpsFlowDefinition("builderWithWorkspaceRequirement()", true));
+        r.assertLogContains(MissingContextVariableException.class.getCanonicalName(), r.buildAndAssertStatus(Result.FAILURE, p));
+    }
+
+    public static class BuilderWithoutWorkspaceRequirement extends Builder implements SimpleBuildStep {
+        @DataBoundConstructor
+        public BuilderWithoutWorkspaceRequirement() {
+        }
+        // TODO: Remove once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 250:
```suggestion
        // TODO: Remove once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -203,4 +203,88 @@ public void builderWithEnvironment() throws Exception {
...
+        public void perform(@Nonnull Run<?, ?> run, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull TaskListener listener) throws InterruptedException, IOException {
+            fail("This method should not get called.");
+        }
+        // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 254:
```suggestion
        // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -203,4 +203,88 @@ public void builderWithEnvironment() throws Exception {
...
+        public void perform(@Nonnull Run<?, ?> run, @Nonnull EnvVars env, @Nonnull TaskListener listener) throws InterruptedException, IOException {
+            listener.getLogger().println("workspace context not needed.");
+        }
+        // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 258:
```suggestion
        // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreStepTest.java
```diff
@@ -203,4 +203,88 @@ public void builderWithEnvironment() throws Exception {
...
+        public void perform(@Nonnull Run<?, ?> run, @Nonnull FilePath workspace, @Nonnull EnvVars env, @Nonnull Launcher launcher, @Nonnull TaskListener listener) throws InterruptedException, IOException {
+            listener.getLogger().println("workspace context not needed, but provided.");
+        }
+        // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 262:
```suggestion
        // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -286,4 +287,109 @@ public static Slave createSpecialEnvSlave(JenkinsRule rule, String nodeName, @Ch
         }
     }
 
+    public static final class WrapperWithWorkspaceRequirement extends SimpleBuildWrapper {
+        @DataBoundConstructor public WrapperWithWorkspaceRequirement() { }
+        @Override public void setUp(@Nonnull Context context, @Nonnull Run<?, ?> build, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull TaskListener listener, @Nonnull EnvVars initialEnvironment) throws IOException, InterruptedException {
+            listener.getLogger().println(">>> workspace context required and provided.");
+            context.setDisposer(new DisposerWithWorkspaceRequirement());
+        }
+        // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 296:
```suggestion
        // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -286,4 +287,109 @@ public static Slave createSpecialEnvSlave(JenkinsRule rule, String nodeName, @Ch
...
+        public void setUp(@Nonnull Context context, @Nonnull Run<?, ?> build, @Nonnull TaskListener listener, @Nonnull EnvVars initialEnvironment) throws IOException, InterruptedException {
+            listener.getLogger().println(">>> workspace context required but not provided!");
+            context.setDisposer(new DisposerWithWorkspaceRequirement());
+        }
+        public static final class DisposerWithWorkspaceRequirement extends Disposer {
+            @Override public void tearDown(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull TaskListener listener) throws IOException, InterruptedException {
+                listener.getLogger().println("<<< workspace context required and provided.");
+            }
+            // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 305:
```suggestion
            // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -286,4 +287,109 @@ public static Slave createSpecialEnvSlave(JenkinsRule rule, String nodeName, @Ch
...
+                    Run<?, ?> r = story.j.buildAndAssertStatus(Result.FAILURE, p);
+                    story.j.assertLogContains(MissingContextVariableException.class.getCanonicalName(), r);
+                }
+            }
+        });
+    }
+
+    public static final class WrapperWithoutWorkspaceRequirement extends SimpleBuildWrapper {
+        @DataBoundConstructor public WrapperWithoutWorkspaceRequirement() { }
+        // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 342:
```suggestion
        // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -286,4 +287,109 @@ public static Slave createSpecialEnvSlave(JenkinsRule rule, String nodeName, @Ch
...
+        public boolean requiresWorkspace() { return false; }
+        @Override public void setUp(@Nonnull Context context, @Nonnull Run<?, ?> build, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull TaskListener listener, @Nonnull EnvVars initialEnvironment) throws IOException, InterruptedException {
+            listener.getLogger().println(">>> workspace context not needed, but provided.");
+            context.setDisposer(new DisposerWithoutWorkspaceRequirement());
+        }
+        // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 348:
```suggestion
        // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -286,4 +287,109 @@ public static Slave createSpecialEnvSlave(JenkinsRule rule, String nodeName, @Ch
...
+        public void setUp(@Nonnull Context context, @Nonnull Run<?, ?> build, @Nonnull TaskListener listener, @Nonnull EnvVars initialEnvironment) throws IOException, InterruptedException {
+            listener.getLogger().println(">>> workspace context not needed.");
+            context.setDisposer(new DisposerWithoutWorkspaceRequirement());
+        }
+        public static final class DisposerWithoutWorkspaceRequirement extends Disposer {
+            // TODO: Remove once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 354:
```suggestion
            // TODO: Remove once the minimum core version for this plugin is 2.258 or newer.
```

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -286,4 +287,109 @@ public static Slave createSpecialEnvSlave(JenkinsRule rule, String nodeName, @Ch
...
+            public boolean requiresWorkspace() { return false; }
+            @Override public void tearDown(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull TaskListener listener) throws IOException, InterruptedException {
+                listener.getLogger().println("<<< workspace context not needed, but provided.");
+            }
+            // TODO: Mark as @Override once the minimum core version for this plugin is 2.25x or newer.
```
jglick, line 359:
```suggestion
            // TODO: Mark as @Override once the minimum core version for this plugin is 2.258 or newer.
```

jglick (commented):
I am not the maintainer but personally I would advocate just changing

```xml
<jenkins.version>2.258</jenkins.version>
```

and deleting the reflection.

Zastai:
I don't disagree - but this is what I was asked to do, so it's what I did.
I can also see why there might be a desire to depend on LTS releases only.

Dropping the reflection does remove the currently possibility of a plugin being able to get the behaviour without either it or this plugin needing to use Jenkins 2.258. But that's unlikely to be a major factor.

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -286,4 +287,109 @@ public static Slave createSpecialEnvSlave(JenkinsRule rule, String nodeName, @Ch
...
+        @TestExtension("wrapperWithWorkspaceRequirement")
+        public static class DescriptorImpl extends BuildWrapperDescriptor {
+            @Override public boolean isApplicable(AbstractProject<?, ?> project) { return true; }
+        }
+    }
+
+    @Issue("JENKINS-46175")
+    @Test
+    public void wrapperWithWorkspaceRequirement() throws Exception {
+        story.addStep(new Statement() {
```
dwnusbaum, line 320: Prefer `story.then((JenkinsRule foo) -> { ... })` instead in new code. I'll file a PR to mark this method as deprecated in `jenkins-test-harness`.

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
```diff
@@ -92,20 +94,82 @@ public SimpleBuildWrapper getDelegate() {
...
-            BodyInvoker bodyInvoker = getContext().newBodyInvoker();
+            SimpleBuildWrapper.Context c = null;
+            // In Jenkins 2.258, a createContext() method on SimpleBuildWrapper is required to ensure that a Disposer
+            // registered on that context inherits the wrapper's workspace requirement. Use it when available.
+            // TODO: Use 'c = this.delegate.createContext()' once this plugin depends on Jenkins 2.258 or later.
+            try {
+                final Method createContext = SimpleBuildWrapper.class.getMethod("createContext");
+                c = (SimpleBuildWrapper.Context) createContext.invoke(this.delegate);
+            }
+            catch (NoSuchMethodException e) {
```
dwnusbaum, line 105:
```suggestion
            } catch (NoSuchMethodException e) {
```

dwnusbaum (approved):
Thanks! Looks good to me.

src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
dwnusbaum: See https://github.com/jenkinsci/jenkins-test-harness/pull/244.

PR #127, opened 2020-09-21
[JENKINS-29144] 2.249.x dep to avoid reflection & InvocationTargetException

jglick:
#121 caused test failures in `artifact-manager-s3` when on 2.249.x for good reason: `AbortException` was being wrapped in `InvocationTargetException` and thus displaying an ugly stack trace rather than a nice `ERROR` message. Probably `FlowInterruptedException` would have been broken as well, messing up build statuses, termination causes, etc.

Can be corrected by catching `InvocationTargetException` but simpler to just use the LTS dep already and keep the code clean.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.2</version>
+        <version>4.7</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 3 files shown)


Zastai (approved):
LGTM.

Zastai:
I'm assuming this is because any exception leaving the reflection-based `invoke()` gets wrapped?
Then I'll need to update PR #123 to add a catch-and-unwrap to its similar code to deal with the new simple-build-step/wrapper-outside-workspace API. That can then be removed once updating the dep to the next LTS is acceptable.

jglick:
> I'm assuming this is because any exception leaving the reflection-based `invoke()` gets wrapped?

Exactly.

> Then I'll need to update PR #123…

Yes, something like

```java
try {
    perform.invoke(…);
} catch (InvocationTargetException x) {
    Throwable x2 = x.getCause();
    throw x2 instanceof Exception ? (Exception) x2 : x;
}
```

basil (approved):

PR #130, opened 2020-10-01
Add test for timeoutstep with batch script for windows

res0nance:
Adding a test for windows that effectively does the same thing as the linux test


dwnusbaum (approved):
Thanks!

PR #132, opened 2020-10-02
Add Windows tests for CoreWrapperStep

res0nance:
Improves testing for the Windows platform, with this both windows and unix have an equivalent number of tests

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
@@ -99,11 +99,9 @@ public class CoreWrapperStepTest {
                 slaveEnv.put("HOME", "/home/jenkins");
                 createSpecialEnvSlave(story.j, "slave", "", slaveEnv);
                 WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
-                if (Functions.isWindows()) {
-                    p.setDefinition(new CpsFlowDefinition("node('slave') {mock {semaphore 'restarting'; echo \"groovy PATH=${env.PATH}:\"; bat 'echo shell PATH=%PATH%:'}}", true));
-                } else {
-                    p.setDefinition(new CpsFlowDefinition("node('slave') {mock {semaphore 'restarting'; echo \"groovy PATH=${env.PATH}:\"; sh 'echo shell PATH=$PATH:'}}", true));
-                }
+                p.setDefinition(new CpsFlowDefinition("node('slave') {mock {semaphore 'restarting'; echo \"groovy PATH=${env.PATH}:\"; " + (Functions.isWindows()
+                        ? "bat 'echo shell PATH=%PATH%:'}}"
+                        : "sh 'echo shell PATH=$PATH:'}}"), true));
                 WorkflowRun b = p.scheduleBuild2(0).getStartCondition().get();
                 SemaphoreStep.waitForStart("restarting/1", b);
             }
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -90,13 +90,20 @@
...
+                }
                 slaveEnv.put("HOME", "/home/jenkins");
                 createSpecialEnvSlave(story.j, "slave", "", slaveEnv);
                 WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
-                p.setDefinition(new CpsFlowDefinition("node('slave') {mock {semaphore 'restarting'; echo \"groovy PATH=${env.PATH}:\"; sh 'echo shell PATH=$PATH:'}}", true));
+                if (Functions.isWindows()) {
+                    p.setDefinition(new CpsFlowDefinition("node('slave') {mock {semaphore 'restarting'; echo \"groovy PATH=${env.PATH}:\"; bat 'echo shell PATH=%PATH%:'}}", true));
+                } else {
+                    p.setDefinition(new CpsFlowDefinition("node('slave') {mock {semaphore 'restarting'; echo \"groovy PATH=${env.PATH}:\"; sh 'echo shell PATH=$PATH:'}}", true));
+                }
```
dwnusbaum, line 106: nit: I would reduce the duplication here a bit:

```suggestion
                p.setDefinition(new CpsFlowDefinition("node('slave') {mock {semaphore 'restarting'; echo \"groovy PATH=${env.PATH}:\"; " + (Functions.isWindows()
                        ? "bat 'echo shell PATH=%PATH%:'}}"
                        : "sh 'echo shell PATH=$PATH:'}}"), true));
```

dwnusbaum (approved):
Thanks!

dwnusbaum:
There were some infra issues in the CI build, will try rebuilding.

> EC2 (aws) - Windows 2019 (i-0e380e8dd44d9e228) was marked offline: Connection was broken: java.io.EOFException

dwnusbaum:
Hmm, `TimeoutStepTest.activityRemote` failed in the CI build on the Windows branch. The result was `SUCCESS` instead of `ABORTED` and `ShouldNot` was printed to the logs even though it should not have been. @res0nance you just added support for Windows to that test in https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/130, but if it is flaky, maybe we should remove the Windows support again? I am not sure if it is an intermittent timing problem, a Jenkins core problem with killing processes on Windows, a problem with the `timeout` step, or something else.

res0nance:
I'll take a look at that soon and report what I find with it

res0nance:
Hey Devin 👋 

I've taken a look at this flake and while I cannot reproduce this in its current form I have a theory on how this occurs, AFAICT its an intermittent timing issue.

Given a working run here
```
  15.815 [p #1] [Pipeline] {
  15.815 [p #1] [Pipeline] timeout
  15.815 [p #1] Timeout set to expire after 5 sec without activity
  15.815 [p #1] [Pipeline] {
  15.878 [p #1] [Pipeline] bat
  16.573 [p #1] NotHere
  18.461 [p #1] NotHereYet
  20.410 [p #1] JustHere
  26.486 [p #1] Cancelling nested steps due to timeout
  26.486 [p #1] Sending interrupt signal to process
  28.604 [p #1] ^CTerminate batch job (Y/N)?
  28.605 [p #1] ^Cscript returned exit code -1073741510
  29.045 [p #1] [Pipeline] }
  29.112 [id=75]        INFO    o.j.p.workflow.job.WorkflowRun#finish: p #1 completed: ABORTED
  29.115 [p #1] [Pipeline] // timeout
  29.115 [p #1] [Pipeline] }
  29.115 [p #1] [Pipeline] // node
```
We see that from the time we send the interrupt signal to the time the batch execution gets interrupted is a whole 2 seconds.
Now when I reduce the time before Should not is printed via `ping -n 10 127.0.0.1 >NUL` setting this to 8 causes a failure on my system.

From what I see we have 2 ways of moving forward either we disable the test on windows or we increase the timeout to hopefully prevent this flake on Windows. WDYT?

dwnusbaum:
@res0nance Thanks for looking into it! I think we can try increasing the timeout on Windows for now, but if it flakes after that, let's just remove the Windows support from the test.

dwnusbaum:
Rebuilding

PR #134, opened 2020-10-23
Disposer.requiresWorkspace is final

jglick:
#123 mistakenly kept an override of `Disposer.requiresWorkspace`, which in the ultimate version of the API was not necessary (this method need only be overridden in `SimpleBuildWrapper`), and is in fact `final`. Because of reflection, this mistake was not noticed until PCT uncovered it in https://github.com/jenkinsci/bom/pull/339.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
@@ -115,8 +115,6 @@ public class CoreWrapperStep extends Step {
                 if (workspace != null && launcher != null) {
                     this.delegate.setUp(c, run, workspace, launcher, listener, env);
                 } else {
-                    // If we get here, workspaceRequired is false and there is no workspace context. In that case, the
-                    // overload of setUp() introduced in Jenkins 2.258 MUST exist.
                     delegate.setUp(c, run, listener, env);
                 }
             }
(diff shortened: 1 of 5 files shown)


timja (approved):

Zastai (approved):
Not sure it makes sense to do this without also changing the other TODOs related to 2.258, but I have nothing against it either.

Zastai:
Looks like with this change, the test fails on Jenkins < 2.258 (not unexpected) because SimpleBuildWrapper always requires a workspace there.

jglick:
A follow up can bump the baseline to 2.263 and delete reflection, but I thought I would file the minimal patch needed for PCT.

jglick:
> the test fails on Jenkins < 2.258

Hmm because of https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/af1995070706a86da15b53f23871839c00375b82/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java#L230. So the attempt to make this system usable by plugins with an older core dep never actually worked.

jglick:
Time to get rid of reflection. I am just going to update the baseline already.

src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
```diff
@@ -147,16 +117,7 @@ private void doStart() throws Exception {
                 } else {
                     // If we get here, workspaceRequired is false and there is no workspace context. In that case, the
                     // overload of setUp() introduced in Jenkins 2.258 MUST exist.
```
Zastai, line 119: This comment can go as well (it was dropped in CoreStep.java (line 118) too).

Zastai (approved):
LGTM

dwnusbaum:
`ReadWriteFileStepTest.testBinaryFileRoundtrip` failed on Windows with `CompositeIOException: Unable to delete 'C:\Users\jenkins\Work\workspace\rkflow-basic-steps-plugin_PR-134\target\tmp\j h6257972212631507846'. Tried 3 times (of a maximum of 3) waiting 0.1 sec between attempts.`, will try rebuilding.

dwnusbaum (approved):

PR #135, opened 2020-11-06
Reduce flakiness of windows timeout test

res0nance:
Increase the time sleep before the next print.
This should never fail as it likely will be force killed after 5 seconds.

See https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/132#issuecomment-723038487


dwnusbaum (approved):
Thanks! Discussed in https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/132#issuecomment-722465147 and later comments.

PR #137, opened 2020-11-13
Corrected flakes in StashTest

jglick:
Seen for example in https://github.com/jenkinsci/bom/pull/357. As can be seen in https://github.com/jenkinsci/workflow-job-plugin/blob/32090697af62e9513de78dd1ed096a6bace8651e/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L639-L643 the build is marked completed just before stashes are cleared.


bitwiseman (approved):

jglick:
Do you plan to merge this?

basil:
@car-roll This is needed for BOM as well

car-roll (approved):

PR #102, opened 2019-12-11
[JENKINS-60354] Use FlowInterruptedException.isActualInterruption to decide if an exception should be rethrown

dwnusbaum:
See [JENKINS-60354](https://issues.jenkins-ci.org/browse/JENKINS-60354) and https://github.com/jenkinsci/workflow-step-api-plugin/pull/51.

PR #144, opened 2021-05-29
[JENKINS-51454] Pipeline retry operation doesn't retry when there is a timeout inside of it

basil:
See [JENKINS-51454](https://issues.jenkins.io/browse/JENKINS-51454). Downstream of jenkinsci/workflow-step-api-plugin#61.

### Problem

There are two desired behaviors depending on the placement of `timeout` and `retry` relative to each other.

#### First case ([JENKINS-51454](https://issues.jenkins.io/browse/JENKINS-51454)):

```groovy
retry(3) {
  timeout(time: 5, unit: 'MINUTES') {
    // Something that can fail
  }
}
```

In this case, if the timeout triggers, the desired behavior is for `retry` to run its body again. This is _not_ the current behavior.

#### Second case ([JENKINS-44379](https://issues.jenkins.io/browse/JENKINS-44379)):

```groovy
timeout(time: 5, unit: 'MINUTES') {
  retry(3) {
    // Something that can fail
  }
}
```

In this case, if the timeout triggers, the desired behavior is for `retry` to be aborted without retrying anything. This is the current behavior, which is working as-designed after [JENKINS-44379](https://issues.jenkins.io/browse/JENKINS-44379).

### Solution

[JENKINS-60354](https://issues.jenkins.io/browse/JENKINS-60354) overlaps a bit with this issue. @dwnusbaum suggested that the approach from #81 by @rudolfwalter could be tweaked a bit to use jenkinsci/workflow-step-api-plugin#51 (if a setter was added) to change the value of `FlowInterruptedException#actualInterruption` from `true` to `false` when the exception propagates past the timeout step that threw the exception, in which case the fix for [JENKINS-60354](https://issues.jenkins.io/browse/JENKINS-60354) in #102 would fix [JENKINS-51454](https://issues.jenkins.io/browse/JENKINS-51454) as well. That is precisely what we are doing in jenkinsci/workflow-step-api-plugin#61 and this PR.


src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java
```diff
@@ -64,6 +65,11 @@ public void onFailure(StepContext context, Throwable t) {
                     TaskListener l = context.get(TaskListener.class);
                     if (t instanceof AbortException) {
                         l.error(t.getMessage());
+                    } else if (t instanceof FlowInterruptedException) {
+                        FlowInterruptedException fie = (FlowInterruptedException) t;
+                        for (CauseOfInterruption cause : fie.getCauses()) {
+                            l.getLogger().println(cause.getShortDescription());
```
jglick, line 71:
```suggestion
                            cause.print(l);
```

is better as the impl can include hyperlinks and the like.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -233,13 +247,70 @@ private void cancel() {
...
         private static final long serialVersionUID = 1L;
 
+        /** Flow node ID of the corresponding timeout block */
+        private String nodeId;
+
+        private String getNodeId() {
+            return nodeId;
+        }
+
+        private void setNodeId(String nodeId) {
```
jglick, line 310: Would make code simpler to just move this into the constructor (and make the field `final`).

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -233,13 +247,70 @@ private void cancel() {
 
     }
 
+    private class Callback2 extends BodyExecutionCallback {
+
+        @Override
+        public void onSuccess(StepContext context, Object result) {
+            if (killer != null) {
+                killer.cancel(true);
```
jglick, line 255: I assume is unlikely to throw exceptions? See special behaviors lost when moving from `TailCall`. Maybe better to enhance `TailCall` to make `onFailure` non`final` so you can call `super` but still do some special processing with the exception.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -162,7 +169,10 @@ private void cancel() {
                     LOGGER.log(Level.WARNING, null, x);
                     return;
                 }
-                final Throwable death = new FlowInterruptedException(Result.ABORTED, new ExceededTimeout());
+                ExceededTimeout exceededTimeout = new ExceededTimeout();
+                exceededTimeout.setNodeId(nodeId);
+                final Throwable death =
+                        new FlowInterruptedException(Result.ABORTED, exceededTimeout);
```
jglick, line 175: I am not understanding why all the complexity. Does a simple
```suggestion
                        new FlowInterruptedException(Result.ABORTED, false, exceededTimeout);
```
not solve the issue here, without the upstream PR?

dwnusbaum: If you do that, then I think both case 1 and case 2 from the PR description would have the same behavior. (Maybe fine, but IMO that just trades the current frustrations with case 1 to new frustrations with case 2.)

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
basil: Done. I wasn't sure whether to put the call to `super.onFailure()` before or after the new logic, but I put it before the new logic and the tests still passed.

timja:
@dwnusbaum @car-roll

basil:
The test that just failed is flaky. I just re-ran it locally and it passed.

basil:
```
12:14:18  Cannot contact aci-maven-11-2thrx: hudson.remoting.ChannelClosedException: Channel "hudson.remoting.Channel@68f899af:JNLP4-connect connection from 40.88.240.16/40.88.240.16:6273": Remote call on JNLP4-connect connection from 40.88.240.16/40.88.240.16:6273 failed. The channel is closing down or has closed down
```

I am stepping away from this for now. I believe the change is ready for merge and release, but the CI build is failing due to problems that I believe are out of my control and not caused by me.

timja:
Build is green now, cc @dwnusbaum

car-roll (approved):

car-roll:
@dwnusbaum would you mind taking a look at this since @jglick is not available atm?

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -256,6 +263,34 @@ private static ExecutorService newExecutorService() {
             }
         }
 
+        @Override
+        public void onFailure(StepContext context, Throwable t) {
+            super.onFailure(context, t);
```
dwnusbaum, line 268: I think it would make more sense to call this at the end of the method rather than the start so that the exception is updated before `context.onFailure(t)` is called in the superclass. I guess it's fine since this code occupies the CPS VM thread so other steps will not run until everything returns, but it seems confusing.

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
```diff
@@ -179,4 +179,50 @@ public void downstreamBuildFailureShouldRetry() throws Exception {
...
     }
+
+    @Issue("JENKINS-44379")
+    @Test
+    public void shouldNotRetryAfterOuterTimeout() throws Exception {
+        WorkflowJob p = r.createProject(WorkflowJob.class);
+        p.setDefinition(
+                new CpsFlowDefinition(
+                        "int count = 0\n"
+                                + "timeout(time: 5, unit: 'SECONDS') {\n"
```
dwnusbaum, line 190: nit: The indentation here and in the next test is weird. I prefer to use trailing `+` operators for Pipeline scripts in tests so that the script all starts at the same level of indentation.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -256,6 +263,34 @@ private static ExecutorService newExecutorService() {
...
+            if (t instanceof FlowInterruptedException) {
```
dwnusbaum, line 269: Might be worth adding a system property to revert to the old behavior in case it causes problems for some users. For example, something like this:
```suggestion
            if (t instanceof FlowInterruptedException && !Boolean.getBoolean(TimeoutStepExecution.class.getName() + ".forceInterruption")) {
```

If you do this, I would mention the property in the release notes.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -256,6 +263,34 @@ private static ExecutorService newExecutorService() {
...
+                    flowNode = getContext().get(FlowNode.class);
+                } catch (IOException | InterruptedException e) {
+                    LOGGER.log(Level.WARNING, null, e);
+                }
+                if (flowNode != null) {
+                    String nodeId = flowNode.getId();
+                    for (CauseOfInterruption cause : ((FlowInterruptedException) t).getCauses()) {
+                        if (cause instanceof ExceededTimeout) {
+                            ExceededTimeout exceededTimeout = (ExceededTimeout) cause;
+                            if (nodeId.equals(exceededTimeout.getNodeId())) {
```
dwnusbaum, line 285: nit: Maybe worth updating the comment on `ExceededTimeout.nodeId` to mention that `exceededTimeout.getNodeId()` could be `null` for exceptions serialized before this change. Everything should work fine though since you're not using it as the receiver here, so it would just match the behavior from before this change.

dwnusbaum (approved):
Seems fine, but I added some comments. The main thing I would change would be to call `FlowInterruptedException.handle` instead of directly printing the causes.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
dwnusbaum: Oh, I just saw https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/144#discussion_r643216989. Like I mentioned above, I don't think it matter because of the CPS VM thread, but I'm not 100% sure. If there is an issue, I guess it would be some kind of race condition where an enclosing step sees the failure before the exception gets updated.

basil: I moved it to the end as you suggested.

src/test/java/org/jenkinsci/plugins/workflow/steps/RetryStepTest.java
basil: I started formatting my code with `google-java-format` a long time ago in order to avoid having to debate formatting preferences.

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
basil: Done

src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
basil: Done, modulo the release note update, which I assume Carroll would take care of as the one doing the release.

basil:
All feedback has been addressed, and the build is green. I think it's time to put this bug to rest.

rajarajeshwarie:
does the timeout be enabled in the retry of the pipeline? I tried the code..when the timeout exceeded, the retry option was working but without timeout.

stage (‘Build’) {

        options {
        retry(2)
        timeout(time: 10, unit: 'MINUTES')
        
        }
        steps {
           //code
        }
        post {  
            success { 
             //code
            }
        }
    }

jglick:
Check in Scripted syntax. If that works, and you only see a problem in Declarative syntax, file a bug report in `pipeline-model-definition-plugin`.

PR #145, opened 2021-05-31
Migrate from `RestartableJenkinsRule` to `JenkinsSessionRule`

basil:
We seem to be generally preferring `JenkinsSessionRule` to `RestartableJenkinsRule`. The main benefit to `JenkinsSessionRule` is that `#then` runs immediately, so it plays nicely with things like `@After.`


jglick (approved):

offa:
This gets rid of calls to deprecated `addStep()` too.

PR #147, opened 2021-08-01
Make `workflow-basic-steps` compatible with Guava 21.0 and newer

basil:
Discovered in jenkinsci/bom#423. Like jenkinsci/workflow-support-plugin#114 and jenkinsci/git-client-plugin#686, but for `workflow-basic-steps`.

`workflow-basic-steps` previously used `MoreExecutors#sameThreadExecutor` from Guava, which was removed in version 21.0 in favor of `MoreExecutors#newDirectExecutorService`. This PR uses reflection to first try to call `MoreExecutors#newDirectExecutorService` but then call `MoreExecutors#sameThreadExecutor` if the first method is not available to support both older and newer versions of Guava.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
@@ -17,9 +17,12 @@ import java.io.OutputStream;
 import java.io.Serializable;
 import java.lang.ref.Reference;
 import java.lang.ref.WeakReference;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


timja:
@jenkinsci/workflow-basic-steps-plugin-developers please

timja:
@basil you may want to pull in https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/149 as well, incremental doesn't work otherwise in bom for PCT

jglick (approved):
Probably OK though https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/0016470142333097412fc3c797f422b92f9315ed/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java#L173

timja:
@car-roll

car-roll (approved):

timja:
@car-roll is it possible to get a release please?

Ideally including https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/144

car-roll:
@timja no problem, once i've finished going through #144, and the other approvals clear, I'll do a release

PR #166, opened 2022-01-08
Bump `mock-javamail` from 1.9 to 1.12

basil:
See [the release notes](https://github.com/jenkinsci/lib-mock-javamail/releases/tag/mock-javamail-1.12).

PR #167, opened 2022-01-08
Use detached JavaMail plugin

basil:
Downstream of #166 and jenkinsci/jenkins#6165. Bumps the baseline to one containing jenkinsci/jenkins#6165, thus using the detached JavaMail plugin, but does not upgrade to Jakarta Mail.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.25</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.263.1</jenkins.version>
+        <jenkins.version>2.329-rc31978.fa_f50a_a_d3e46</jenkins.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/6165 -->
         <java.level>8</java.level>
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -143,6 +143,12 @@
             <artifactId>javadoc</artifactId>
             <version>1.6</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
```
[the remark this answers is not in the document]
basil: Nope

basil:
I've marked this as ready for review because the PR is complete, but I'm not in any rush for this to be merged due to the bump in Jenkins baseline. I just wanted to file the PR to close the loop on remaining tasks regarding detaching JavaMail.

jglick (approved):

car-roll (approved):

pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.25</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.263.1</jenkins.version>
+        <jenkins.version>2.331</jenkins.version>
```
olamy, line 67: we should refrain using non LTS this prevents users use last version of the plugin....

jglick: So? If and when there is some important feature/fix that cannot wait for the next LTS line, https://gist.github.com/jglick/86a30894446ed38f918050c1180483e2 is easy enough.

olamy: why not but it's not something so simple and probably not obvious for the majority of the plugins maintainers.
probably worth some documentation somewhere.
how do that work with CD? 
does the cd tooling can detect he needs to generate something between 23343_kljffe0345 and 53343_kljloe0345?

jglick: > probably worth some documentation somewhere.

Agreed.

> how do that work with CD?

The same Gist shows an example. Again, yes, deserves formal documentation and perhaps some tooling to make it easier and more discoverable.

PR #168, opened 2022-01-08
Migrate to Jakarta Mail

basil:
Downstream of #166, #167, and jenkinsci/jenkins#6165. Bumps the baseline to one containing jenkinsci/jenkins#6165 and migrates to Jakarta Mail.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,6 +88,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>mailer</artifactId>
+            <version>410.va_7940d9ded39</version> <!-- TODO https://github.com/jenkinsci/mailer-plugin/pull/144 -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -161,12 +162,12 @@
             <scope>test</scope>
         </dependency>
         <!--
-            This must come before javax-mail-api in the class path in order to avoid eclipse-ee4j/mail#350.
+            This must come before jakarta-mail-api in the class path in order to avoid eclipse-ee4j/mail#350.
         -->
         <dependency>
             <groupId>org.jvnet.mock-javamail</groupId>
             <artifactId>mock-javamail</artifactId>
-            <version>1.12</version>
+            <version>2.0</version>
             <scope>test</scope>
         </dependency>
         <!--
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -88,6 +88,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>mailer</artifactId>
+            <version>403.v4da_0b_7596fa_6</version> <!-- TODO https://github.com/jenkinsci/mailer-plugin/pull/144 -->
```
jglick, line 91: https://github.com/jenkinsci/mailer-plugin/pull/144

basil:
I think this should be released to avoid problems for people who have upgraded to Mailer 435.v79ef3972b_5c7 or later (which uses `jakarta.mail` imports).

basil:
See [the release notes for Email Extension 2.90](https://github.com/jenkinsci/email-ext-plugin/releases/tag/email-ext-2.90) for suggested messaging.

jglick:
> I think this should be released

Done: https://github.com/jenkinsci/workflow-basic-steps-plugin/releases/tag/980.v82219a_ed188e

> avoid problems for people who have upgraded to Mailer 435.v79ef3972b_5c7 or later

Is there some specific sort of problem you would foresee?

> [the release notes for Email Extension 2.90](https://github.com/jenkinsci/email-ext-plugin/releases/tag/email-ext-2.90)

That seems to be specific to people using custom in-process Groovy scripting, which would not apply here.

Maybe that the signature of `MimeMessageBuilder` changed incompatibly?

basil:
> Is there some specific sort of problem you would foresee?

Yes, I suspect the `mail` step of Pipeline: Basic Steps 975.v8d89b_e9579b_b_ would stop working on 2.332.x once Mailer is upgraded to 435.v79ef3972b_5c7, because Pipeline: Basic Steps 975.v8d89b_e9579b_b_ uses JavaMail and consumes this library from Mailer, which (as of 435.v79ef3972b_5c7) is providing that library under new imports, which Pipeline: Basic Steps doesn't consume until 980.v82219a_ed188e.

> That seems to be specific to people using custom in-process Groovy scripting, which would not apply here.

What do you mean by "that"?

The "Jakarta imports" section is specific to people using custom Groovy scripting, which does not apply here.

The "Related plugin releases" section does apply here.

jglick:
> The "Related plugin releases" section does apply here.

OK, I updated https://github.com/jenkinsci/mailer-plugin/releases/tag/435.v79ef3972b_5c7 and https://github.com/jenkinsci/workflow-basic-steps-plugin/releases/tag/980.v82219a_ed188e accordingly; can you check for accuracy please?

basil:
Looks good to me

jglick:
OK. I am preparing a `bom` update for this…

basil:
Beat you to it in https://github.com/jenkinsci/bom/pull/1280

PR #170, opened 2022-01-13
Remove "Checks if running on a Unix-like node"

SaranshIam:
Removed the method that prints "Checks if running on a Unix-like node".
We see a lot of noise in the output because of this statement as it is called every time the docker-workflow-plugin and the hashicorp-vault-plugin uses the isUnix() method.
I also see a ticket filed on Jenkins to remove this (https://issues.jenkins.io/browse/JENKINS-61195), So this will be beneficial for many other consumers.

- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ x] Please describe what you did
- [ x] Link to relevant issues in GitHub or Jira
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/IsUnixStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/IsUnixStep.java
@@ -61,9 +61,6 @@ public class IsUnixStep extends Step {
             return "isUnix";
         }
 
-        @Override public String getDisplayName() {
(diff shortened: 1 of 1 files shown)


wimleers:
Can you post before/after screenshots?

jglick (changes requested):
This is not an appropriate fix.

PR #180, opened 2022-03-02
document environment variables are case insensitive

jtnord:
Document that environment variables are case insensitive but case preserving as per https://javadoc.jenkins.io/hudson/EnvVars.html#:~:text=So%20for%20a%20consistent%20cross%20platform%20behavior%2C%20it%20creates%20the%20least%20confusion%20to%20make%20the%20table%20case%20insensitive%20but%20case%20preserving.

- [ ] Link to relevant issues in GitHub or Jira
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/steps/EnvStep/help.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/steps/EnvStep/help.html
@@ -1,6 +1,6 @@
 <div>
 Sets one or more environment variables within a block.
-The names of environment variables are case insensitive but case preserving, i.e., setting `Foo` will change the value of `FOO` if it already exists.
+The names of environment variables are case-insensitive but case-preserving, that is, setting `Foo` will change the value of `FOO` if it already exists.
 Environment variables are available to any external processes spawned within that scope.
 For example:
 <p><pre>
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/steps/EnvStep/help.html
```diff
@@ -1,6 +1,7 @@
 <div>
 Sets one or more environment variables within a block.
-These are available to any external processes spawned within that scope.
+The names of environment variables are case insensitive but case preserving.
```
daniel-beck, line 3:
```suggestion
The names of environment variables are case insensitive but case preserving, i.e., setting `Foo` will change the value of `FOO` if it already exists.
```

alecharp (approved):

src/main/resources/org/jenkinsci/plugins/workflow/steps/EnvStep/help.html
```diff
@@ -1,6 +1,7 @@
 <div>
 Sets one or more environment variables within a block.
-These are available to any external processes spawned within that scope.
+The names of environment variables are case insensitive but case preserving, i.e., setting `Foo` will change the value of `FOO` if it already exists.
```
jtnord, line 3:
```suggestion
The names of environment variables are case insensitive but case preserving, that is, setting `Foo` will change the value of `FOO` if it already exists.
```

ammanders: Yes, but case-insensitive and case-preserving should be hyphenated. Thanks!

car-roll (approved):

PR #195, opened 2022-05-02
[JENKINS-49707] Tests of `ErrorCondition` across restarts

jglick:
Downstream of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180. ~Subsumes #203~, mostly just adding test coverage.


src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -106,6 +111,7 @@ public class CoreWrapperStepTest {
                 SemaphoreStep.waitForStart("restarting/1", b);
         });
         sessions.then(j -> {
+                j.waitOnline((Slave) j.jenkins.getNode("slave")); // TODO otherwise sh will fail until the agent reconnects
```
jglick, line 114: It is possible that the `SecretMasker` portion of https://github.com/jenkinsci/kubernetes-plugin/pull/1083/commits/2141a03fe3c1920171a26182910de3b79e41960d means that https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180/commits/b65ff1cd10a77700c017ac5bf660b75de1eb8c56 could finally be reverted. Working around the test failure for now but would prefer to do it right.

src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStep.java
```diff
@@ -39,6 +42,7 @@
 public class RetryStep extends Step {
     
     private final int count;
+    private List<ErrorCondition> errorConditions;
```
jglick, line 45: Perhaps should be called simply `conditions`.

jglick:
#219 and #220 contain everything of value left in here.

PR #258, opened 2023-02-16
Moved `PushdStep` to `workflow-durable-task-step`

jglick:
Downstream of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/293.


src/main/java/org/jenkinsci/plugins/workflow/steps/PushdStep.java
```diff
@@ -24,84 +24,22 @@
...
-            FilePath dir = getContext().get(FilePath.class).child(path);
-            getContext().get(TaskListener.class).getLogger().println("Running in " + dir);
-            getContext().newBodyInvoker()
-                    // TODO perhaps should just be moved into workflow-durable-task-step?
-                    .withContext(FilePathDynamicContext.createContextualObject(dir))
-                    // Could use a dedicated BodyExecutionCallback here if we wished to print a message at the end ("Returning to ${cwd}"):
-                    .withCallback(BodyExecutionCallback.wrap(getContext()))
-                    .start();
-            return false;
+            throw new AssertionError("only deserialized");
```
dwnusbaum, line 42: Did you test deserialization of the old execution with this PR?

dwnusbaum (approved):

src/main/java/org/jenkinsci/plugins/workflow/steps/PushdStep.java
jglick: Yes, see testing notes in upstream.

dwnusbaum: Ok great; It looks like I was reviewing the upstream PR at the same time you were adding the testing notes, so I did not see them before.

PR #265, opened 2023-03-22
Update parent

jglick:
Supersedes #264 & #261, and part of #263.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.54</version>
+        <version>4.56</version>
         <relativePath/>
     </parent>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -165,6 +160,12 @@
             <artifactId>mock-javamail</artifactId>
             <version>2.0</version>
             <scope>test</scope>
+            <exclusions>
```
raul-arabaolaza, line 163: Nit: A note explaining why would be interesting

raul-arabaolaza (approved):

pom.xml
jglick: This is a pretty standard exclusion used in various plugins IIRC. I am a bit hesitant to leave comments with too many details with such cases because if you doubt that an `exclusion` is still actually necessary, it is best to just try deleting it and running

```bash
mvn clean verify -Dtest=InjectedTest
```

PR #308, opened 2024-04-24
Log a note when `catchError` sets build result

jglick:
Otherwise it can be confusing why a build had the result that it did. Observed in a pipeline using the idiom

```groovy
retry(count: 2, conditions: [kubernetesAgent(), nonresumable()]) {
  node(POD_LABEL) {
    stage('build') {
      withMockLoad(averageDuration: 180) {
        catchError(buildResult: 'UNSTABLE') {
          sh MOCK_LOAD_COMMAND
        }
      }
    }
    stage('publish') {
      junit 'mock-junit.xml'
      archiveArtifacts artifacts: 'mock-artifact-*.txt', allowEmptyArchive: true, fingerprint: true
    }
  }
}
```

When the first pod is killed, the `node` block retried and the build ran to completion (`FlowEndNode` set to `SUCCESS`) yet the `Run` is set to `UNSTABLE`, which is what the step is supposed to do but unintentional here; the reason for the `catchError` was rather to work around an exit code of 1 from the mock load command due to mock test failures, which is better handled via https://github.com/jenkinsci/mock-load-builder-plugin/pull/102.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStep.java
@@ -248,6 +248,7 @@ public final class CatchErrorStep extends Step implements CatchExecutionOptions 
                         Functions.printStackTrace(t, listener.getLogger());
                     }
                     if (buildResult.isWorseThan(Result.SUCCESS)) {
+                        listener.getLogger().println("Setting overall build result to " + buildResult);
                         context.get(Run.class).setResult(buildResult);
                     }
                     if (stepResult.isWorseThan(Result.SUCCESS)) {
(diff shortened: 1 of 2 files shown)


julieheard (approved):

dwnusbaum (approved):

src/main/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStep.java
```diff
@@ -248,6 +248,7 @@ public Object readResolve() {
                         Functions.printStackTrace(t, listener.getLogger());
                     }
                     if (buildResult.isWorseThan(Result.SUCCESS)) {
+                        listener.getLogger().println("Setting overall build result to " + buildResult);
```
jtnord, line 251: shouldn't this only log if the **current** result is worse than this?

e.g. the result is currently `FAILURE`  and buildResult is `UNSTABLE` it would log "setting build result to UNSTABLE" - but this is incorrect as you can not IIRC improve a build result.

e.g. consider the cut down pipeline:

```
stage('outer  stage') {
    catchError(buildResult: 'UNSTABLE') {
      currentBuild.result = "FAILURE"
      error "nope"
    }
}
```

the result of this pipeline is a FAILURE.  and the buildResult should be FAILURE at the time that catchError is intercepted.   (if you remove the `currentBuild.result = "FAILURE"` then this would infact be unstable.)

jtnord (approved):

PR #321, opened 2025-01-16
Migrate from EE 8 to EE 9

basil:


jglick (approved):

src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -145,6 +146,7 @@ public void killingParallel() throws Throwable {
     @Issue("JENKINS-26521")
     @Test
     public void activity() throws Throwable {
+        assumeThat("TODO consistently failing in ci.jenkins.io yet passing locally", System.getenv("CI"), nullValue());
```
jglick, line 149: (Flakes in these test cases have also been observed in CloudBees CI PCT, though much less frequently.)

PR #347, opened 2025-06-14
Add 'when' step implementation

stuartrowe:
Implement a `when` step so some functionality of the [`when`](https://www.jenkins.io/doc/book/pipeline/syntax/#when) directive available in declarative pipelines is reproduced in scripted pipelines. Specifically tagging a `stage` as skipped when the condition of the `when` step is not met.

This change was motivated by issues using a popular global variable implementation of `when` described in this [SO post](https://stackoverflow.com/questions/56148230/how-to-mark-jenkins-job-stage-as-skipped-for-scripted-pipeline). The problems are discussed [here](https://github.com/jenkinsci/pipeline-graph-view-plugin/issues/504#issuecomment-2964447713) on a bug reported against the [pipeline-graph-view-plugin](https://github.com/jenkinsci/pipeline-graph-view-plugin).

### Testing done

Created new tests to demonstrate the functionality of the new step.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,6 +112,10 @@
             <artifactId>apache-httpcomponents-client-4-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-stage-step</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
             <scope>test</scope>
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/WhenStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/WhenStep.java
@@ -1,0 +1,188 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2025 Stuart Rowe
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.steps;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
+import hudson.model.Run;
+import java.io.IOException;
+import java.util.AbstractCollection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import org.jenkinsci.plugins.workflow.actions.LabelAction;
+import org.jenkinsci.plugins.workflow.actions.TagsAction;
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -111,6 +111,10 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>apache-httpcomponents-client-4-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-stage-step</artifactId>
```
stuartrowe, line 116: It may make more sense to implement the `when` step in the [pipeline-stage-step-plugin](https://github.com/jenkinsci/pipeline-stage-step-plugin)?

stuartrowe:
@jenkinsci/workflow-basic-steps-plugin-developers could someone please review this change?

stuartrowe:
@jglick @dwnusbaum @basil is anyone able to review this change please?

pom.xml
jglick: Definitely this dependency may not be here.

src/main/java/org/jenkinsci/plugins/workflow/steps/WhenStep.java
```diff
@@ -0,0 +1,189 @@
...
+    // from org.jenkinsci.plugins.pipeline.StageStatus
+    private static final String TAG_STAGE_STATUS = "STAGE_STATUS";
+    private static final String SKIPPED_STAGE_STATUS = "SKIPPED_FOR_CONDITIONAL";
+
+    private boolean condition;
+
+    @DataBoundConstructor
+    public WhenStep(Object condition) {
+        // Support for a reasonable subset of expressions following the rules of Groovy Truth (see
+        // https://groovy-lang.org/semantics.html#the-groovy-truth)
```
jglick, line 59: Please, no. Take a `boolean`.

src/main/java/org/jenkinsci/plugins/workflow/steps/WhenStep.java
```diff
@@ -0,0 +1,189 @@
...
+import org.kohsuke.stapler.DataBoundConstructor;
+
+/**
+ * A conditional block scoped step for a stage
+ */
+public class WhenStep extends Step {
+
+    // from org.jenkinsci.plugins.pipeline.StageStatus
+    private static final String TAG_STAGE_STATUS = "STAGE_STATUS";
+    private static final String SKIPPED_STAGE_STATUS = "SKIPPED_FOR_CONDITIONAL";
```
jglick, line 52: Like, from Declarative? Should stay there IMO, not be ported to Scripted.

jglick (changes requested):
Design does not feel right at all. Prefer simply

```groovy
if (someCondition()) {
  stage('whatever') {
    sh 'stuff'
  }
}
```

If there is a compelling need for some Pipeline visualizations to actually display the stage, but greyed out or something, then it should be a `boolean` option on `stage` to make such a notation (and then decline to run its block):

```groovy
stage(name: 'whatever', skipped: !someCondition()) {
  sh 'stuff'
}
```

with a corresponding change to `pipeline-model-definition-plugin` to use the proper step rather than hacky hard-coded stage tags, and matching changes to visualizations, which really requires a dedicated `Action` for `FlowNode`s defined in `workflow-api`.

dwnusbaum:
> which really requires a dedicated `Action` for `FlowNode`s defined in workflow-api.

In this case you might be able to reuse [`NotExecutedNodeAction`](https://github.com/jenkinsci/workflow-api-plugin/blob/f682ae82e57b81b26cf81bec60f34c0a9552e08a/src/main/java/org/jenkinsci/plugins/workflow/actions/NotExecutedNodeAction.java), which is currently used for Declarative's "Restart from stage" and CloudBees' "Restart from Checkpoint" to mark nodes before the restart point. This seems close enough to what Declarative is currently doing with `when` and its custom "SKIPPED_FOR_CONDITIONAL" tag that these could probably be used for the same purpose.

> with a corresponding change to pipeline-model-definition-plugin to use the proper step 

I don't think this would be trivial. Declarative has options to check `when` before or after various other steps, see https://www.jenkins.io/doc/book/pipeline/syntax/#evaluating-when-before-entering-agent-in-a-stage (see [this logic](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/56a15e805f12dd0afa412693f5179a9beda1326f/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy#L262-L308)), which would make it difficult for the plugin to adopt a feature that is tied to the `stage` step. I don't think this is a blocker though; it is probably good enough just to have Declarative use a standard `Action` for reporting stages that it marks as skipped.

jglick:
> Declarative has options to check `when` before or after various other steps…

Do not know any details of that, but I thought I had: maybe we could just throw out the explicit marking of a skipped stage altogether, and simply have visualizations treat a `stage` (or I guess `parallel` branch) as “skipped” if it completes with no nested steps, or perhaps only `echo` steps? So you could write simply

```groovy
stage('whatever') {
  if (someCondition()) {
    sh 'stuff'
  }
}
```

or as desired

```groovy
stage('whatever') {
  if (someCondition()) {
    sh 'stuff'
  } else {
    echo 'Condition was not met this time'
  }
}
```

and visualizations would work naturally.

